PR #11, opened 2023-11-22
DS-172 Add RadioButton with temp CSS

jeremy-davis-sonarsource:


src/components/radio-button-group/RadioButtonGroup.tsx
```diff
@@ -0,0 +1,154 @@
...
+  value?: string;
+}
+
+export function RadioButtonGroup(props: Readonly<Props>) {
+  const { onChange, options, ...groupProps } = props;
+
+  return (
+    <RadioGroupWrapper onValueChange={onChange} {...groupProps}>
+      {options.map((o) => (
+        <RadioButton groupId={groupProps.id} {...o} />
```
gregaubert, line 43: We probably want the option id as a key too here right?

src/components/radio-button-group/RadioButtonGroup.tsx
```diff
@@ -0,0 +1,154 @@
...
+ * If the label is a string, we use it directly as a fallback if no aria-label is provided
+ */
+type RadioButtonProps = {
+  disabled?: boolean;
+  groupId?: string;
+  value: string;
+} & ({ ariaLabel: string; label: ReactNode } | { ariaLabel?: string; label: string });
+
+function RadioButton(props: Readonly<RadioButtonProps>) {
+  const { ariaLabel, disabled, groupId = 'radio', label, value } = props;
```
gregaubert, line 60: Should we always have a default groupId?  or just make the group id non optional?

src/components/radio-button-group/RadioButtonGroup.tsx
```diff
@@ -0,0 +1,154 @@
...
+  position: relative;
+
+  &::after {
+    content: '';
+    display: block;
+
+    box-sizing: border-box;
+    height: var(--radio-size-indicator);
+    width: var(--radio-size-indicator);
+    border-radius: 50%;
```
gregaubert, line 136: Do we have design token for radius-full ? we probably should

src/components/radio-button-group/RadioButtonGroup.tsx
```diff
@@ -0,0 +1,154 @@
...
+
+    background-color: var(--radio-fg);
+  }
+
+  &[data-disabled]::after {
```
gregaubert, line 141: This data-disabled thing is the way Radix handle disabled state? Or some custom stuff on our side?

.storybook/preview.tsx
```diff
@@ -2,9 +2,16 @@ import { ThemeProvider } from '@emotion/react';
...
+import '../src/generated/design-tokens-light.css';
+import '../src/generated/design-tokens-dark.css';
 
 const GlobalStoriesStyle = styled.div`
-  color: blue;
+  // color: blue;
+
+  * {
+    all: unset;
+  }
```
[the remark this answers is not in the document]
jeremy-davis-sonarsource: I think we need to be exhaustive actually... otherwise ew risk having an impact on the consumer app

package.json
```diff
@@ -56,5 +56,8 @@
     "typescript": "5.2.2",
     "vite": "5.0.0",
     "vite-plugin-dts": "3.6.3"
+  },
+  "dependencies": {
+    "@radix-ui/react-radio-group": "1.1.3"
```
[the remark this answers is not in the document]
jeremy-davis-sonarsource: I don't think so. Dependencies like this one are distributed, so I think it makes sense to make it explicit.

src/components/radio-button-group/RadioButtonGroup.tsx
jeremy-davis-sonarsource: Yep, missed it for some reason!

src/components/radio-button-group/RadioButtonGroup.tsx
jeremy-davis-sonarsource: Good point, let's make it mandatory

src/components/radio-button-group/RadioButtonGroup.tsx
jeremy-davis-sonarsource: Yes, that's radix: https://www.radix-ui.com/primitives/docs/components/radio-group#indicator

src/components/radio-button-group/RadioButtonGroup.tsx
jeremy-davis-sonarsource: Yeah, let's keep that as a fix for when we have all the right tokens for the component

stories/RadioButtonGroup-stories.tsx
```diff
@@ -0,0 +1,52 @@
...
+import { RadioButtonGroup } from '@sonarsource/echoes-react';
+
+const meta = {
+  component: RadioButtonGroup,
+  title: 'RadioButtonGroup',
+} satisfies Meta<typeof RadioButtonGroup>;
+
+export default meta;
+
+type Story = StoryObj<typeof RadioButtonGroup>;
```
gregaubert, line 11: Seems like this can be `typeof meta` when using the satisfies keyword, makes for less things to replace when copy pasting stories :P

gregaubert (approved):
Looks good! 👍🏻

PR #12, opened 2023-11-22
Add layer 1 with base tokens and colors

gregaubert:


design-tokens/tokens/layer1/base.json
```diff
@@ -113,78 +125,150 @@
...
-      "value": "{fontSizes.body} * 1",
+    "headingS": {
+      "value": "0.75rem",
       "type": "fontSizes"
     },
-    "body": {
-      "value": "16",
+    "bodyLarge": {
+      "value": "1rem",
       "type": "fontSizes"
```
jeremy-davis-sonarsource, line 191: Maybe a stupid question, but shouldn't `1 rem` be the "normal" size?

design-tokens/tokens/layer1/base.json
```diff
@@ -1,109 +1,121 @@
...
     "heading": {
-      "value": "110%",
+      "value": "132%",
       "type": "lineHeights"
     },
     "body": {
-      "value": "140%",
+      "value": "142%",
       "type": "lineHeights"
     }
```
jeremy-davis-sonarsource, line 120: NP: Shouldn't these be in `rem`?

design-tokens/tokens/layer1/base.json
```diff
@@ -113,78 +125,150 @@
...
+      "value": "0.75rem",
       "type": "fontSizes"
     }
+  },
+  "borderWidth": {
+    "default": {
+      "value": "0.063rem",
+      "type": "borderWidth"
+    }
+  },
```
jeremy-davis-sonarsource, line 207: Is it a good idea to tie the border width to the `rem`?
If I change the font size, I don't expect the borders to grow as well, I guess

design-tokens/tokens/layer2/base.json
```diff
@@ -4,22 +4,22 @@
       "Bold": {
         "type": "typography",
         "value": {
-          "fontFamily": "{fontFamilies.heading}",
-          "fontWeight": "{fontWeights.headingBold}",
+          "fontFamily": "{fontFamilies.mono}",
```
jeremy-davis-sonarsource, line 7: I think we want `sans` here. `mono` is for code snippets

jeremy-davis-sonarsource (commented):
Not sure if I'm asking for changes, or just asking questions :shrug:

design-tokens/tokens/layer1/base.json
```diff
@@ -1,109 +1,121 @@
...
-    "heading": {
-      "value": "Inter",
+    "mono": {
+      "value": "Ubuntu Mono",
       "type": "fontFamilies"
     },
-    "body": {
-      "value": "Roboto",
+    "sans": {
+      "value": "Inter",
```
gregaubert, line 108: We need fallback font families, I suggest we use the same as tailwind for that:
https://tailwindcss.com/docs/font-family

design-tokens/tokens/layer1/base.json
gregaubert: Agree, that's also currently the case in MIUI

design-tokens/tokens/layer1/base.json
```diff
@@ -113,78 +125,150 @@
...
+    "bodyMedium": {
+      "value": "0.875rem",
       "type": "fontSizes"
     },
-    "xs": {
-      "value": "{fontSizes.body} * 0.65",
+    "bodySmall": {
+      "value": "0.75rem",
       "type": "fontSizes"
     }
```
gregaubert, line 200: We shouldn't have a 0.75rem, it's 12px it's too small, we dropped 12px font size everywhere with the MIUI effort we shouldn't go back to it.
Smallest should be 14px IMO.

Same applies fro headingS by the way.

design-tokens/tokens/layer1/base.json
gregaubert: Agree also, seems like the rem value is trying to reach for a 1px value but can't really.

design-tokens/tokens/layer1/base.json
```diff
@@ -1,109 +1,121 @@
...
       "type": "fontFamilies"
     },
-    "body": {
-      "value": "Roboto",
+    "sans": {
+      "value": "Inter",
       "type": "fontFamilies"
     }
   },
   "lineHeights": {
```
gregaubert, line 112: Let's keep singular for all the categories

design-tokens/tokens/layer1/base.json
```diff
@@ -113,78 +125,150 @@
...
       "type": "paragraphSpacing"
     },
-    "h2": {
-      "value": "26",
+    "large": {
+      "value": "1.375rem",
       "type": "paragraphSpacing"
     }
   },
   "fontWeights": {
```
gregaubert, line 150: Let's keep singular for all the categories

design-tokens/tokens/layer1/base.json
```diff
@@ -113,78 +125,150 @@
...
       "type": "fontWeights"
     },
-    "bodyBold": {
-      "value": "Bold",
+    "regular": {
+      "value": "Regular",
       "type": "fontWeights"
     }
   },
   "fontSizes": {
```
gregaubert, line 164: Let's keep singular for all the categories

design-tokens/tokens/layer1/colors.json
```diff
@@ -1,392 +1,404 @@
...
-        "value": "#e2e8f0",
-        "type": "color"
-      },
-      "400": {
-        "value": "#cbd5e0",
-        "type": "color"
-      },
-      "500": {
-        "value": "#a0aec0",
+    "support": {
```
gregaubert, line 3: Should we add transparent to support colors?

design-tokens/tokens/layer1/base.json
marciopmoreira6: Discussed and reviewed at the office with @gregaubert. We are making the conscious decision to keep 12px which should only be used when displaying meta-data. Documentation to support this design decision will follow so we can ensure that this value is not used wrongfully like it was in the past.

design-tokens/tokens/layer1/base.json
marciopmoreira6: 16 is the base that we used to calculate the type scale. Is the boundary between a text and a heading. With 16 you can go up and down, to find other font sizes.

"Default" or "normal" means that is the most used value, the go-to font size and that value is 14px. 
- If you need to display relevant information, you use Default. 
- If you need less relevant info, like metadata, you use the small.
- If you need to display impactful information, you go for the large.

design-tokens/tokens/layer1/base.json
marciopmoreira6: Good point. I'm not sure.
The goal here is that these 2 values will be the multipliers that you can use to create the line height. So, if your font size is 1rem, the line height is going to be (1rem * 142% = 22,72). What value should it be in rem that allows us to reach the same effect?

design-tokens/tokens/layer1/base.json
marciopmoreira6: Good catch. Will update

design-tokens/tokens/layer1/base.json
marciopmoreira6: Thanks. Ill add those Tailwind values there

design-tokens/tokens/layer1/colors.json
marciopmoreira6: We can. This would be something like:
**Layer 1**
support - transparent
**Layer 2**
bg--none  - transparent

Right?

design-tokens/tokens/layer1/base.json
gregaubert: And actually when looking at the current UI most of the text is 14px

design-tokens/tokens/layer1/base.json
gregaubert: Ah interesting, that's a good point. Let's keep it in %

design-tokens/tokens/layer1/base.json
gregaubert: In our last discussion didn't you say you couldn't do it in figma?

design-tokens/tokens/layer1/colors.json
gregaubert: yeah looks good

design-tokens/tokens/layer1/base.json
marciopmoreira6: Yes. This token cannot be used in Figma. Figma has a fallback font built-in. But I can add it as a string so you guys can use it in code.

gregaubert: Ok but that's gonna be problematic if you can't use it in figma right?
I think we are better off adding it only on our side directly.

design-tokens/tokens/layer1/base.json
marciopmoreira6: True :)

design-tokens/tokens/layer1/base.json
marciopmoreira6: No, it's ok. Don't worry.

design-tokens/tokens/layer1/base.json
```diff
@@ -1,109 +1,129 @@
...
-      "lg": {
-        "value": "{size.dimension.md} * {size.dimension.scale}",
+      "100": {
+        "value": "0.5rem",
         "type": "dimension"
       },
-      "xl": {
-        "value": "{size.dimension.lg} * {size.dimension.scale}",
+      "125": {
+        "value": "0.75rem",
```
gregaubert, line 25: This one should be "150" to respect the percentage scale you told me about

design-tokens/tokens/layer1/base.json
```diff
@@ -1,109 +1,129 @@
...
-        "value": "{size.dimension.xs}",
-        "type": "spacing"
       },
-      "sm": {
-        "value": "{size.dimension.sm}",
-        "type": "spacing"
+      "150": {
+        "value": "1.25rem",
+        "type": "dimension"
       },
```
gregaubert, line 31: This one should be "250" to respect the percentage scale you told me about

design-tokens/tokens/layer1/base.json
```diff
@@ -1,109 +1,129 @@
...
+          "type": "dimension"
+        },
+        "md": {
+          "value": "1.25rem",
+          "type": "dimension"
+        },
+        "lg": {
+          "value": "1.5rem",
+          "type": "dimension"
+        }
```
gregaubert, line 76: Currently, our icons can only be 1rem right? Do we intend to change that?

design-tokens/tokens/layer1/base.json
```diff
@@ -1,109 +1,129 @@
...
       },
-      "xl": {
-        "value": "16",
-        "type": "borderRadius"
+      "800": {
+        "value": "4rem",
+        "type": "dimension"
+      }
+    },
+    "sizes": {
```
gregaubert, line 57: Singular?

design-tokens/tokens/layer1/base.json
```diff
@@ -1,109 +1,129 @@
...
+        "xs": {
+          "value": "0.25rem",
+          "type": "dimension"
+        }
       },
-      "multi-value": {
-        "value": "{size.borderRadius.sm} {size.borderRadius.lg}",
-        "type": "borderRadius",
-        "description": "You can have multiple values in a single radius token. Read more on these: https://docs.tokens.studio/available-tokens/border-radius-tokens#single--multiple-values"
+      "icons": {
```
gregaubert, line 64: Singular?

design-tokens/tokens/layer1/base.json
```diff
@@ -1,109 +1,129 @@
...
+          "type": "dimension"
+        }
       }
     }
   },
-  "opacity": {
-    "low": {
-      "value": "10%",
-      "type": "opacity"
+  "borderRadius": {
```
gregaubert, line 80: borderRaidus are in rem in the current tailwind config, should it be the same here?
Or should it be relative to the element it's applied to?

sm: 2 (px?) is probably small for something that has 100px size but it's huge for something that has a 4px size.

So either it should be relative or the name should not convey a meaning of small, medium, large and just be numbers like for our spacings.

design-tokens/tokens/layer1/base.json
```diff
@@ -1,109 +1,129 @@
...
-    "body": {
-      "value": "Roboto",
+    "fallbackMono": {
+      "value": "ui-monospace, SFMono-Regular, Menlo, Monaco, Consolas, \"Liberation Mono\", \"Courier New\", monospace",
+      "type": "fontFamilies"
+    },
+    "fallbackSans": {
+      "value": "ui-sans-serif, system-ui, -apple-system, BlinkMacSystemFont, \"Segoe UI\", Roboto, \"Helvetica Neue\", Arial, \"Noto Sans\", sans-serif, \"Apple Color Emoji\", \"Segoe UI Emoji\", \"Segoe UI Symbol\", \"Noto Color Emoji\"",
       "type": "fontFamilies"
     }
```
gregaubert, line 118: Ok I see, we will have to see how we can work with that in the lib and the typography components. There might be some changes in the future.

design-tokens/tokens/layer1/base.json
```diff
@@ -113,78 +133,150 @@
...
+      "value": "Regular",
       "type": "fontWeights"
     }
   },
-  "fontSizes": {
-    "h1": {
-      "value": "{fontSizes.h2} * 1.25",
+  "fontSize": {
+    "headingXXXL": {
+      "value": "2.125rem",
```
gregaubert, line 174: Current biggest in MIUI is 2.25rem do we want to align?

design-tokens/tokens/layer1/base.json
```diff
@@ -113,78 +133,150 @@
...
+    "bodySmall": {
+      "value": "0.75rem",
       "type": "fontSizes"
     }
+  },
+  "borderWidth": {
+    "default": {
+      "value": "1px",
+      "type": "borderWidth"
+    }
```
gregaubert, line 214: In current MIUI we also have some bigger border width:
default: ['1px',
stroke: ['2px',
active: ['3px',
focus: ['4px',

design-tokens/tokens/layer1/base.json
```diff
@@ -113,78 +133,150 @@
...
+  },
+  "paragraph": {
```
gregaubert, line 216: Do we also want typographies for headings ?

Currently in MIUI we have:

```
    '.heading-xl': {
      'font-family': theme('fontFamily.sans'),
      'font-size': theme('fontSize.xl'),
      'line-height': theme('fontSize').xl[1],
      'font-weight': theme('fontWeight.semibold'),
    },
    '.heading-lg': {
      'font-family': theme('fontFamily.sans'),
      'font-size': theme('fontSize.lg'),
      'line-height': theme('fontSize').lg[1],
      'font-weight': theme('fontWeight.semibold'),
    },
    '.heading-md': {
      'font-family': theme('fontFamily.sans'),
      'font-size': theme('fontSize.md'),
      'line-height': theme('fontSize').md[1],
      'font-weight': theme('fontWeight.semibold'),
    },
```

gregaubert: We are also missing typography for code:

```
    '.code': {
      'font-family': theme('fontFamily.mono'),
      'font-size': theme('fontSize.code'),
      'line-height': theme('fontSize').code[1],
      'font-weight': theme('fontWeight.regular'),
    },
    '.code-highlight': {
      'font-family': theme('fontFamily.mono'),
      'font-size': theme('fontSize.code'),
      'line-height': theme('fontSize').code[1],
      'font-weight': theme('fontWeight.bold'),
    },
    '.code-comment': {
      'font-family': theme('fontFamily.mono'),
      'font-size': theme('fontSize.code'),
      'line-height': theme('fontSize').code[1],
      'font-style': 'italic',
    },
```

design-tokens/tokens/layer1/base.json
```diff
@@ -113,78 +133,150 @@
...
+        "fontFamily": "{fontFamily.sans}",
+        "fontWeight": "{fontWeight.regular}",
+        "lineHeight": "{lineHeight.body}",
+        "fontSize": "{fontSize.bodyLarge}",
+        "letterSpacing": "{letterSpacing.default}",
+        "paragraphSpacing": "{paragraphSpacing.large}"
+      },
+      "type": "typography"
+    },
+    "defaultSemibold": {
```
gregaubert, line 250: If we have that for default should we also have it for others ? small and large?
Or should we just get rid of that and when we change the fontweight we just apply the needed fontweight to any typography token ?

design-tokens/tokens/layer1/base.json
marciopmoreira6: Someone clearly failed at math 😆  Thanks, updated

design-tokens/tokens/layer1/base.json
marciopmoreira6: Thanks, Updated

design-tokens/tokens/layer1/base.json
marciopmoreira6: Yes. An icon size is defined by the line height of a text. If you are using text-large (fs: 16/ lh: ~24) you should pair it with icon-large (h: 24px).

design-tokens/tokens/layer1/base.json
marciopmoreira6: This is based on what a revamp we did to our typescale. The one I shared for review with you @gregaubert.

design-tokens/tokens/layer1/base.json
marciopmoreira6: great point. I'll update this to relative values.

marciopmoreira6: Actually, I was reading more about it, and I think we should keep the static pixel values, and instead change the token name to an abstract value.

design-tokens/tokens/layer2/dark.json
```diff
@@ -1,76 +1,316 @@
...
+        "accent": {
+          "default": {
+            "accent": {
+              "type": "color",
+              "value": "{color.indigo.200}"
+            },
+            "accent-hover": {
+              "type": "color",
+              "value": "{color.indigo.300}"
+            },
```
gregaubert, line 54: variable names are not gonna make any sense:
`--echoes-color-background-accent-default-accent`

We need to remove the first accent object or remove the second `accent` in the name.

By the way, this is the same for all the following design tokens like weak-weak, warning-default-warning, success-weak-weak, danger-..., info-...

design-tokens/tokens/layer2/dark.json
```diff
@@ -1,76 +1,316 @@
...
+              "type": "color",
+              "value": "#5D6CD033"
+            },
+            "weak-hover": {
+              "type": "color",
+              "value": "#5D6CD04D"
+            },
+            "weak-active": {
+              "type": "color",
+              "value": "#5D6CD04D"
```
gregaubert, line 75: Why do we have raw colors ? Layer 2 should always reference a layer 1 color.

design-tokens/tokens/layer2/dark.json
```diff
@@ -1,76 +1,316 @@
...
+          "default": {
+            "warning": {
+              "type": "color",
+              "value": "{color.yellow.200}"
+            }
+          },
+          "weak": {
+            "weak": {
+              "type": "color",
+              "value": "#FADC7933"
```
gregaubert, line 93: Why do we have raw colors ? Layer 2 should always reference a layer 1 color.

design-tokens/tokens/layer2/dark.json
```diff
@@ -1,76 +1,316 @@
...
+          "default": {
+            "success": {
+              "type": "color",
+              "value": "{color.yellowGreen.200}"
+            }
+          },
+          "weak": {
+            "weak": {
+              "type": "color",
+              "value": "#C5E67C33"
```
gregaubert, line 107: Why do we have raw colors ? Layer 2 should always reference a layer 1 color.

design-tokens/tokens/layer2/dark.json
```diff
@@ -1,76 +1,316 @@
...
+            },
+            "danger-disabled": {
+              "type": "color",
+              "value": "{color.grey.500}"
+            }
+          },
+          "weak": {
+            "weak": {
+              "type": "color",
+              "value": "#FDA29B33"
```
gregaubert, line 137: Why do we have raw colors ? Layer 2 should always reference a layer 1 color.

design-tokens/tokens/layer2/dark.json
```diff
@@ -1,76 +1,316 @@
...
+          "default": {
+            "info": {
+              "type": "color",
+              "value": "{color.blue.200}"
+            }
+          },
+          "weak": {
+            "weak": {
+              "type": "color",
+              "value": "#8FCAEA33"
```
gregaubert, line 163: Why do we have raw colors ? Layer 2 should always reference a layer 1 color.

design-tokens/tokens/layer2/light.json
gregaubert, line 1: Every comment I made on dark.json also apply to light.json

gregaubert:
@marciopmoreira6 @jeremy-davis-sonarsource I think we are good to go to merge the current version. I'll need a thumbs-up though to merge.
Once merge I can fix generate the CSS variables and fix the radiobutton to use these tokens.

jeremy-davis-sonarsource (approved):

PR #13, opened 2023-11-23
Update design tokens generation script and add theme support

gregaubert:


design-tokens/build.js
```diff
@@ -1,93 +1,130 @@
...
 
-StyleDictionary.registerTransformGroup({
-  name: CUSTOM_TRANSFORM_GROUP,
-  transforms: ['attribute/cti', ...transforms, 'name/cti/kebab'],
-});
+initStyleDictionary();
+buildBaseTokens();
+buildThemedTokens($themes);
+buildCSSRootFile($themes);
+buildThemesEnumType($themes);
```
jeremy-davis-sonarsource, line 19: You're redeclaring this `$themes` variable in the functions that use it (the argument), which might lead to weird bugs in the future.
How about creating an extra function to wrap that part?

```
function readAndBuildThemes() {
  const themes = parse(read())

  buildThemedTokens(themes);
  buildCSSRootFile(themes);
  buildThemesEnumType(themes);
}
```

NP: rename it `themeDefinitions`, to clarify what it is?

jeremy-davis-sonarsource (changes requested):
Nice work, I've just got one remark

jeremy-davis-sonarsource:
I also noticed the ticket number is wrong on this commit: [DS-178 Add theme support in storybook](https://github.com/SonarSource/echoes-react/pull/13/commits/56058422cba1c33792828ac4dd47b02cc850235c)

design-tokens/build.js
gregaubert: ah yeah good point! 👍🏻

gregaubert: I didn't add an extra function but renamed to something clearer and to avoid the naming collision

PR #14, opened 2023-11-23
DS-178 Add jest and RTL

jeremy-davis-sonarsource:
Only the last commit is part of this


src/components/radio-button-group/__tests__/RadioButtonGroup-test.tsx
```diff
@@ -0,0 +1,23 @@
+import { render, screen } from '@testing-library/react';
```
gregaubert, line 1: Let's keep in mind to create a test utils file in the near future to not access the render function directly, we can have the userEvent setup done automatically as part of it too.

src/components/radio-button-group/__tests__/RadioButtonGroup-test.tsx
```diff
@@ -0,0 +1,23 @@
...
+import { RadioButtonGroup } from '../RadioButtonGroup';
```
gregaubert, line 2: I'm wondering if we should also have the tests running against the built components ?

jest.config.cjs
```diff
@@ -0,0 +1,51 @@
...
+        outputDirectory: 'test-results/',
+        outputName: 'junit.xml',
+        ancestorSeparator: ' > ',
+        suiteNameTemplate: '{filename}',
+        classNameTemplate: '{classname}',
+        titleTemplate: '{title}',
+      },
+    ],
+  ],
+  testTimeout: 60000,
```
gregaubert, line 50: Let's keep it low until it's really a problem ?

jest.config.cjs
```diff
@@ -0,0 +1,51 @@
...
+          target: 'es2018',
+        },
+      },
+    ],
+  },
+  transformIgnorePatterns: ['/node_modules/(?!(d3-.+))/'],
+  reporters: [
+    'default',
+    [
+      'jest-junit',
```
gregaubert, line 39: What's the advantage of the junit reporter ? Makes me tickle to see java related stuff on this repo

jest.config.cjs
```diff
@@ -0,0 +1,51 @@
+module.exports = {
+  coverageDirectory: '<rootDir>/test-results/coverage',
+  collectCoverageFrom: ['src/**/*.{ts,tsx}', '!src/**/*-stories.tsx', '!src/**/index.ts'],
+  coverageReporters: ['lcovonly', 'text'],
+  moduleFileExtensions: ['ts', 'tsx', 'js'],
+  moduleNameMapper: {
+    '^.+\\.(md|jpg|jpeg|png|gif|eot|otf|webp|svg|ttf|woff|woff2|mp4|webm|wav|mp3|m4a|aac|oga)$':
+      '<rootDir>/config/jest/FileStub.js',
+    '^.+\\.css$': '<rootDir>/config/jest/CSSStub.js',
```
gregaubert, line 9: I think we are missing those two stub files

jest.config.cjs
```diff
@@ -0,0 +1,51 @@
...
+  setupFiles: ['<rootDir>/config/jest/SetupTestEnvironment.ts'],
+  setupFilesAfterEnv: [
+    '<rootDir>/config/jest/SetupReactTestingLibrary.ts',
+    '<rootDir>/config/jest/SetupJestAxe.ts',
+  ],
+  snapshotSerializers: ['@emotion/jest/serializer'],
+  testEnvironment: 'jsdom',
+  testPathIgnorePatterns: [
+    '<rootDir>/config',
+    '<rootDir>/design-system',
```
gregaubert, line 20: Should be design-tokens folder instead, also generated and dist folders

config/jest/SetupReactTestingLibrary.ts
gregaubert, line 1: I would see this as a specific setup for the CI on longer term and therefore maybe rename it, our tests shouldnt be so slow locally

jest.config.cjs
```diff
@@ -0,0 +1,51 @@
...
+    '<rootDir>/node_modules',
+    '<rootDir>/scripts',
+  ],
+  testRegex: '(/__tests__/.*|\\-test)\\.(ts|tsx)$',
```
gregaubert, line 24: We also need to update the vite config to ignore test files and make sur we don't end up with some tests in the build.

jest.config.cjs
```diff
@@ -0,0 +1,51 @@
...
+  transform: {
+    '^.+\\.(t|j)sx?$': [
+      '@swc/jest',
```
gregaubert, line 27: Can't we leverage vite for the transform ? To avoid adding more technologies than needed in the repo

jest.config.cjs
```diff
@@ -0,0 +1,51 @@
...
+      {
+        jsc: {
+          target: 'es2018',
+        },
+      },
+    ],
+  },
+  transformIgnorePatterns: ['/node_modules/(?!(d3-.+))/'],
```
gregaubert, line 35: I don't think this is gonna work since we don't have a node modules folder

gregaubert (commented):
I only did a code review, didn't test, I think the main stuff to address is the ignore/coverage folders in the jest config and the vite config to ignore test files.

Otherwise most of my other comments are longer term when we put the CI in place.

jest.config.cjs
jeremy-davis-sonarsource: we can probably remove it and use the default, for now

jest.config.cjs
jeremy-davis-sonarsource: isn't it required for SQ consumption?

jest.config.cjs
jeremy-davis-sonarsource: Oh, we can probably remove those... we don't import any from the components

jest.config.cjs
jeremy-davis-sonarsource: I'm not sure, but you can try!

jest.config.cjs
gregaubert: Ah I'm not sure then, SC doesn't need it that's for sure

jest.config.cjs
gregaubert: Just had a quick look and it seems like vite integration with jest isn't so great, let's keep it like this.

jest.config.cjs
gregaubert: I didn't find anything in SQ doc talking about jest-junit, I'll drop it for now and if it's needed when we put the CI in place we can add it back

gregaubert (approved):

PR #15, opened 2023-11-24
DS-177 Enforce standard license header on all files and in dist folder

gregaubert:


package.json
```diff
@@ -18,11 +18,13 @@
...
     "build": "vite build",
     "build-tokens": "node design-tokens/build.js",
-    "ts-check": "tsc --noEmit",
-    "storybook": "storybook dev -p 6006",
     "build-storybook": "storybook build",
     "format": "prettier --write --list-different .",
-    "format-check": "prettier --check ."
+    "format-check": "prettier --check .",
+    "lint": "eslint --ext js,ts,tsx,snap --quiet .",
+    "lint-report": "eslint --ext js,ts,tsx,snap -f json -o build-report/eslint-report.json .",
```
jeremy-davis-sonarsource, line 25: Should be `build-reports`

src/components/radio-button-group/RadioButtonGroup.tsx
```diff
@@ -18,18 +18,18 @@
...
   // Group Props
   defaultValue?: string;
   disabled?: boolean;
+
   id: string;
+  onChange?: (value: string) => void;
+  options: RadioOption[];
   required?: boolean;
   value?: string;
 }
```
jeremy-davis-sonarsource, line 35: This breaks the semantic separation that I introduced. Maybe we shouldn't enforce prop ordering?

src/generated/design-tokens-light.css
```diff
@@ -22,11 +22,4 @@
   --card-border-radius: 8px;
   --card-padding: 16px;
   --card-background: #ffffff;
-
-
-  /* Test for RadioButton */
-  --radio-fg: #5D6CD0;
-  --radio-bg: #E8EBFF;
-  --radio-fg-disabled: #C5CDDF;
-  --radio-bg-disabled: #EFF2F9;
```
jeremy-davis-sonarsource, line 31: I know this is just paperclips and duct-tape, but I think we should keep it until we get the actual tokens

jeremy-davis-sonarsource (changes requested):
Looks good, although there are a few things that are breaking the RadioButtonGroup as it currently stands

src/components/radio-button-group/RadioButtonGroup.tsx
gregaubert: Ah yes, good catch, I'll drop the rule

src/generated/design-tokens-light.css
gregaubert: Yes I'll add it back to be sure the component continues working, we just need to be careful about this when doing a `yarn build-tokens`

jeremy-davis-sonarsource (approved):

PR #16, opened 2023-11-24
[No Jira] Alignment fix to Radio Button label

jeremy-davis-sonarsource:


src/components/radio-button-group/RadioButtonGroup.tsx
```diff
@@ -152,6 +152,8 @@ const SelectionIndicator = styled(RadioGroup.Indicator)`
 `;
 
 const Label = styled.label`
+  display: block;
```
gregaubert, line 155: I was wondering how you got the alignment right in the demo! I understand now! ahah

I was also wondering if we might need a className prop? and if all our component should make a className prop available? to be able to change stuff like spacing around the component or positioning within flex container, that kind of things.

gregaubert (approved):

PR #17, opened 2023-12-06
DS-172 Fix design tokens usage for RadioButtonGroup component

gregaubert:


design-tokens/tokens/layer3/base.json
```diff
@@ -1 +1,32 @@
...
+          "type": "dimension"
+        },
+        "indicator": {
+          "value": "0.375rem",
+          "type": "dimension"
+        }
+      }
+    }
+  }
+}
```
gregaubert, line 32: @marciopmoreira6 @jeremy-davis-sonarsource I'm not fully confident on this change between layer 1 and 3.
I needed to add some dimensions for the radio button so it seemed logical it should be layer 3, but then I noticed the icons dimensions where in layer 1. I decided to move both in layer 3.
But not sure it's the right approach, I was also wondering if we should have some generic sizes for inputs that would be in layer 1? Also, the fact that it's in layer 3, currently it's gonna be generated twice in the CSS files.

@marciopmoreira6 do you remember why the icons dimension where put in layer 1?

.storybook/preview.tsx
gregaubert, line 1: Added these changes to ensure it's also working well with dark theme in storybook

src/components/radio-button-group/RadioButtonGroup.tsx
```diff
@@ -101,28 +101,28 @@ const Input = styled(RadioGroup.Item)`
...
   box-sizing: border-box;
   padding: 0;
-  border: 1px solid var(--radio-fg);
+  border: 1px solid var(--echoes-color-background-accent-default);
   border-radius: 100%;
 
-  height: var(--size-dimension-md);
-  width: var(--size-dimension-md);
-  min-width: var(--size-dimension-md);
+  height: var(--echoes-radio-button-group-size-input);
```
gregaubert, line 107: Here we might want to have `radiobuttongroup` instead of `radio-button-group` what do you think?

design-tokens/tokens/layer3/base.json
jeremy-davis-sonarsource: I think Icon dimensions can be seen as very fundamental. And radio buttons too.
I guess if we had an "IssueRadioButton" or something with a weird size, that would go into layer 3

gregaubert: mmmh I really don't think a `radiobuttongroup.size.input` value should end up in layer 1, to me that's the definition of a component specific design token.
But if we had a `size.input.small` that we use in `radiobuttongroup` it would make sense for me.

gregaubert: @marciopmoreira6 what's your opinion there?

gregaubert: Following discussions with Marcio we need more thoughts into icons size and how sizes should be handles overall. He will have a look at some other design systems.

In the meantime I'll add a list of sizes inside layer 1 similar to spaces and will use that in radio button for now.

jeremy-davis-sonarsource (approved):
LGTM!

PR #18, opened 2023-12-11
DS-187 Setup react-intl

jeremy-davis-sonarsource:


gregaubert (approved):
Looks good! Let's make sure we don't forget to document how we do translations in components with the default value being normal English

revanshu-paliwal-sonarsource (approved):
LGTM with FV

PR #19, opened 2023-12-11
DS-172 Add a11y check to RadioButtonGroup test

jeremy-davis-sonarsource:


tsconfig.json
```diff
@@ -1,5 +1,5 @@
 {
-  "include": ["src", "stories", "config"],
+  "include": ["src", "stories", "config", "types"],
```
gregaubert, line 2: Should the types folder actually be inside src ?

gregaubert (commented):
One question about types folder location otherwise it looks good

gregaubert (approved):

PR #21, opened 2023-12-11
Fix radio button group tokens based on figma file and update tokens build script

gregaubert:
https://www.figma.com/file/3bo9DCW9JD52Lc21Fcyea8/(Echoes)-Components?type=design&node-id=15-2267&mode=design&t=y4SaQ7XkLllhUgno-4


gregaubert:
@marciopmoreira6 I renamed the base token group to have it all lower case like the other token groups.
Also can you use this branch instead of the one you pushed to check that the changes are fine in figma? I'll drop the old branch you re-created when we were on zoom.

design-tokens/build.js
```diff
@@ -30,14 +30,16 @@ const BASE_TOKENS_NAME = `${NAME_PREFIX}base`;
 const CUSTOM_TRANSFORM_GROUP = 'sonar-design-tokens';
 const CUSTOM_FILTER_NO_COLOR = 'sonar-no-color';
 const CUSTOM_FILTER_THEMED_TOKENS = 'sonar-themed-tokens';
-const themeDefinitions = JSON.parse(fs.readFileSync(`${DESIGN_TOKENS_PATH}/$themes.json`, 'utf-8'));
-const licenceHeader = fs.readFileSync(`config/license/LICENSE-HEADER.txt`, 'utf-8');
+const LICENCE_HEADER = fs.readFileSync(`config/license/LICENSE-HEADER.txt`, 'utf-8');
+
+const TOKEN_GROUPS = JSON.parse(fs.readFileSync(`${DESIGN_TOKENS_PATH}/$themes.json`, 'utf-8'));
+const THEMED_TOKEN_GROUPS = TOKEN_GROUPS.filter(({ group }) => group === 'Themes');
```
jeremy-davis-sonarsource, line 36: NP: I prefer to use Capitalized constants for true hard-coded values. Here we're reading external data, so I think it makes sense to have "normal" constants. Corollary: don't use the `licenseHeader` directly, but pass it as an argument

jeremy-davis-sonarsource (commented):
Just one code-style nitpick, WDYT?

design-tokens/build.js
gregaubert: Ok it make sense

jeremy-davis-sonarsource (approved):
Don't forget to squash!

PR #24, opened 2023-12-12
DS-189 create cirrus pipeline

lucas-paulger-sonarsource:


package.json
```diff
@@ -84,6 +87,7 @@
     "react": "17.0.2",
     "react-dom": "17.0.2",
     "react-intl": "6.5.5",
+    "rollup": "4.8.0",
```
jeremy-davis-sonarsource, line 90: Why do we need this?

jeremy-davis-sonarsource (approved):
Nice work! We can optimize things later on

.cirrus.yml
```diff
@@ -0,0 +1,95 @@
...
+    fingerprint_script: cat yarn.lock
+
+eslint_report_cache_template: &ESLINT_REPORT_CACHE_TEMPLATE
+  eslint_report_cache:
+    folder: 'build-reports/eslint-report'
+    fingerprint_script: echo $CIRRUS_BUILD_ID
+
+jest_coverage_cache_template: &JEST_COVERAGE_CACHE_TEMPLATE
+  jest_coverage_cache:
+    folder: 'build-reports/coverage'
```
jeremy-davis-sonarsource, line 29: So apparently we're missing `/**` at the end of these paths: Ismael just came to my desk to let me know

.cirrus.yml
```diff
@@ -0,0 +1,95 @@
...
+  region: eu-central-1
+  namespace: default
+  docker_arguments:
+    CIRRUS_AWS_ACCOUNT: ${CIRRUS_AWS_ACCOUNT}
+
+yarn_cache_template: &YARN_CACHE_TEMPLATE
+  yarn_cache:
+    folders:
+      - '.yarn'
+    fingerprint_script: cat yarn.lock
```
gregaubert, line 20: so we are going full-on pnp to try with zero install and commit the dependencies? that means no need for this task and the install_dependencies_task?

package.json
lucas-paulger-sonarsource: the install was complaining that it was missing, 
Screenshot 2023-12-13 at 10 47 36

but perhaps it's not needed (confusing why it would complain though) as it mentions if a plugin is written as rollup compatible, it doesn't need rollup. I guess if the license header is working correctly we don't need it 👍🏻 

https://vitejs.dev/guide/api-plugin#rollup-plugin-compatibility

.cirrus.yml
```diff
@@ -0,0 +1,95 @@
...
+  script:
+    - yarn install --immutable
+
+build_number_task:
+  eks_container:
+    <<: *BUILD_TEMPLATE
+    cpu: 1
+    memory: 2Gb
+  script:
+    - .cirrus/build-number.sh
```
gregaubert, line 53: If I understand correctly the build number ends up being `version.build_number` ?

package.json
```diff
@@ -16,16 +16,19 @@
...
     "format": "prettier --write --list-different .",
     "format-check": "prettier --check .",
     "lint": "eslint --ext js,ts,tsx,snap --quiet .",
-    "lint-report": "eslint --ext js,ts,tsx,snap -f json -o build-reports/eslint-report.json .",
+    "lint-report": "eslint --ext js,ts,tsx,snap -f json -o build-reports/eslint-report/eslint-report.json .",
     "storybook": "storybook dev -p 6006",
-    "ts-check": "tsc --noEmit"
+    "ts-check": "tsc --noEmit",
+    "validate": "yarn lint && yarn ts-check && yarn format-check",
+    "validate-ci": "yarn lint-report && yarn ts-check && yarn format-check"
```
gregaubert, line 31: Should we add a check that the design tokens were correctly generated already? Could be another ticket, no need to do it in this PR.

.cirrus.yml
lucas-paulger-sonarsource: It's a non-issue here since we don't have nested folders. everything we cache lives directly inside the `build-reports/coverage` directory:

Screenshot 2023-12-13 at 10 51 12

Of course, verify when we do the scan that there is **some** coverage
Screenshot 2023-12-13 at 10 52 19

lucas-paulger-sonarsource: also we can decide how to change reporting on CI, I didn't create a new `jest.config` for the ci to keep it simple

package.json
```diff
@@ -16,16 +16,19 @@
   "scripts": {
     "dev": "vite build --watch & yarn storybook",
     "test": "jest",
+    "test-ci": "jest --collectCoverage --coverageDirectory=build-reports/coverage",
```
gregaubert, line 19: Let's add a `jest.config.ci` file that extends the normal one with the ci specific config?

package.json
gregaubert: I would say it's worth keeping it since we have some stuff depending on it, instead of having the warning and always coming to questioning ourselves why we see it...

.cirrus.yml
jeremy-davis-sonarsource: as discussed, not an issue

package.json
jeremy-davis-sonarsource: I created a ticket for that :+1:

.cirrus.yml
lucas-paulger-sonarsource: So yes you're right, zero-install includes them
https://github.com/yarnpkg/berry/issues/454#issuecomment-530312089

I looked at yarn itself
https://github.com/yarnpkg/berry/tree/master

it has commits the .yarn/cache and the pnp files. So lets do it!

lucas-paulger-sonarsource: second though will do it as a new PR

gregaubert: yeah make sense!

package.json
lucas-paulger-sonarsource: I would defer until we need to override enough that makes the script too long 🤷🏻

.cirrus.yml
lucas-paulger-sonarsource: There are a few "versions". we will use the one inside the `package.json` and then append the build_number for each build

and call it `Project_Version` (see below) But these aren't actually used yet

https://cirrus-ci.com/task/4769891849338880?logs=main#L50

PR #30, opened 2023-12-13
DS-186 Implement Link component

revanshu-paliwal-sonarsource:


src/components/links/LinkBase.tsx
```diff
@@ -0,0 +1,152 @@
...
+  return (
+    <RouterLink ref={ref} {...rest} onClick={handleClick} to={to}>
+      {icon}
+      {children}
+    </RouterLink>
+  );
+}
+
+const ExternalIcon = styled(OpenNewTabIcon)`
+  color: var(--echoes-color-background-accent-default);
```
gregaubert, line 108: I don't think this is needed, since the color should be following the link color

src/components/links/LinkBase.tsx
```diff
@@ -0,0 +1,152 @@
...
+  margin-left: 4px;
```
gregaubert, line 109: Here you can use the layer 1 design tokens variable: `margin-left: var(--echoes-dimension-space-50)`

src/components/links/LinkBase.tsx
```diff
@@ -0,0 +1,152 @@
...
+`;
+
+export const LinkBase = React.forwardRef(LinkBaseWithRef);
+
+export const LinkBaseStyled = styled(LinkBase)`
```
gregaubert, line 114: Let's extract it to a separate file as we described in the spec document?

src/components/links/LinkBase.tsx
```diff
@@ -0,0 +1,152 @@
...
+
+  &:visited {
+    color: var(--color);
+  }
+
+  &:hover,
+  &:focus,
+  &:active {
+    color: var(--active);
+    text-decoration-color: 1px solid var(--active);
```
gregaubert, line 126: It should only be `text-decoration-color: var(--active)`

src/components/links/LinkBase.tsx
```diff
@@ -0,0 +1,152 @@
...
+
+    svg {
+      color: var(--active);
+    }
+  }
+
+  &:focus,
+  &:focus-visible {
+    background-color: var(--echoes-color-background-accent-weak);
```
gregaubert, line 135: I don't think we want a background color for the focus in this case

src/components/links/LinkBase.tsx
```diff
@@ -0,0 +1,152 @@
...
+      <a
+        rel="noopener noreferrer"
+        target="_blank"
+        {...rest}
+        href={toAsString}
+        onClick={handleClick}
+        ref={ref}>
+        {icon}
+        {children}
+        {hasExternalIcon && <ExternalIcon className="sw-ml-1" />}
```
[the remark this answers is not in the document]
gregaubert: Also, `sw-ml-1` can't be used since it's tailwind, but you already included the spacing in the styled-component so it's fine to just drop it.

src/components/links/LinkBase.tsx
```diff
@@ -0,0 +1,152 @@
...
+
+  &:hover,
+  &:focus,
+  &:active {
+    color: var(--active);
+    text-decoration-color: 1px solid var(--active);
+
+    svg {
+      color: var(--active);
+    }
```
gregaubert, line 130: This shouldn't be needed since the icon follows the color of the text that is already set

src/components/links/LinkBase.tsx
```diff
@@ -0,0 +1,152 @@
...
+
+  ${({ icon }) =>
+    icon &&
+    css`
+      & > svg,
+      & > img {
+        margin-right: 4px;
+      }
+    `};
+`;
```
gregaubert, line 152: We can also add the @media print query from SC to ensure links don't have colors when printed:

```
@media print {
    &,
    &:visited,
    &:hover,
    &:focus,
    &:active {
      color: ${themeColor('currentColor')};
      text-decoration: none;
    }
  }

```

src/components/links/LinkBase.tsx
```diff
@@ -0,0 +1,152 @@
...
+const ExternalIcon = styled(OpenNewTabIcon)`
+  color: var(--echoes-color-background-accent-default);
+  margin-left: 4px;
+`;
+
+export const LinkBase = React.forwardRef(LinkBaseWithRef);
+
+export const LinkBaseStyled = styled(LinkBase)`
+  color: var(--color);
+  text-decoration-color: var(--color);
```
gregaubert, line 116: Let's make sure we are also explicit with the underline, let's add: `text-decoration: solid`

src/components/links/LinkBase.tsx
```diff
@@ -0,0 +1,152 @@
...
+  & > svg {
+    vertical-align: text-bottom !important;
+  }
+
+  ${({ icon }) =>
+    icon &&
+    css`
+      & > svg,
+      & > img {
+        margin-right: 4px;
```
gregaubert, line 149: let's use `var(--echoes-dimension-space-50)`

src/components/links/LinkBase.tsx
```diff
@@ -0,0 +1,152 @@
...
+  text-decoration-color: var(--color);
+
+  &:visited {
+    color: var(--color);
+  }
+
+  &:hover,
+  &:focus,
+  &:active {
+    color: var(--active);
```
gregaubert, line 125: We probably want to make a difference between hover and active since we have two different design tokens for them (with the same color for now), what do you think?

src/components/links/__tests__/Link-test.tsx
```diff
@@ -0,0 +1,138 @@
...
+  const { location } = window;
+  delete (window as { location?: Location }).location;
+  window.location = { ...location, href: '' };
+});
+
+beforeEach(() => {
+  jest.clearAllMocks();
+});
+
+// This functionality won't be needed once we update the breadcrumbs
```
gregaubert, line 37: I wonder if we should keep this comment since we don't have a breadcrumb component

src/components/links/__tests__/Link-test.tsx
```diff
@@ -0,0 +1,138 @@
...
+  await user.click(screen.getByRole('link'));
+
+  expect(screen.getByText('/second')).toBeVisible();
+});
+
+it('external links are indicated by OpenNewTabIcon', () => {
+  setupWithMemoryRouter(<Link to="https://google.com">external link</Link>);
+  expect(screen.getByRole('link')).toBeVisible();
+
+  expect(screen.getByRole('img', { hidden: true })).toBeInTheDocument();
```
gregaubert, line 108: We can remove the `{hidden: true}` and use the toBeVisible once the accessibility is fixed.

src/components/links/__tests__/Link-test.tsx
gregaubert, line 1: Let's make sure we also add tests for:

* the target override
* the noreferrer nofollow logic once it's added
* jest-axe toHaveNoA11yViolations for both external and internal links

stories/LinkComponent-stories.tsx
```diff
@@ -0,0 +1,72 @@
...
+};
+
+export const LinkSubduedWithIcon: StoryFn<typeof LinkSubduedComp> = (args) => (
+  <LinkSubduedComp {...args} />
+);
+LinkSubduedWithIcon.args = {
+  children: 'Link Subdued With Icon',
+  to: '/projects/new',
+  icon: <OpenNewTabIcon />,
+};
```
gregaubert, line 72: Since these are different variants I would move them to their own story file and rewrite them to the new story format.

gregaubert (changes requested):
Good start overall, I've noticed a few things to improve but most comments are minor things.

stories/LinkComponent-stories.tsx
revanshu-paliwal-sonarsource: Will this create a new component for LikSubdued, should it be part of links ?

src/components/links/LinkBase.tsx
revanshu-paliwal-sonarsource: Should the themecolor be replaced with 'currentColor'?

src/components/links/__tests__/Link-test.tsx
```diff
@@ -0,0 +1,138 @@
...
+
+beforeAll(() => {
+  const { location } = window;
+  delete (window as { location?: Location }).location;
+  window.location = { ...location, href: '' };
+});
+
+beforeEach(() => {
+  jest.clearAllMocks();
+});
```
jeremy-davis-sonarsource, line 35: You don't need any of this

stories/LinkComponent-stories.tsx
```diff
@@ -0,0 +1,72 @@
...
+
+type Story = StoryObj<typeof meta>;
+
+export const LinkDefault: Story = {
+  args: {
+    children: 'Link Default',
+    to: '/path',
+  },
+};
+
```
jeremy-davis-sonarsource, line 43: How about stories with some context:

- A link inside a paragraph of text
- A list of links (`ul`)
- a link and other things in a flex-row div with align-items-center

just random ideas, to make sure it behaves nicely in a layout

jeremy-davis-sonarsource (changes requested):
Nice work! Just a couple of things to fix or improve

package.json
```diff
@@ -35,7 +35,8 @@
     "@emotion/styled": "^11.0.0",
     "react": "^17.0.0",
     "react-dom": "^17.0.0",
-    "react-intl": "^6.0.0"
+    "react-intl": "^6.0.0",
+    "react-router-dom": "^6.11.2"
```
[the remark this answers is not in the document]
revanshu-paliwal-sonarsource: This means using 6.0.0 in our dev dependency too, to be in sync ?

src/components/links/LinkBase.tsx
```diff
@@ -0,0 +1,152 @@
...
+        onClick(event);
+      }
+    },
+    [onClick, shouldBlurAfterClick, shouldPreventDefault, shouldStopPropagation, isDisabled],
+  );
+
+  if (isExternal) {
+    return (
+      <a
+        rel="noopener noreferrer"
```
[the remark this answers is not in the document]
revanshu-paliwal-sonarsource: isSonarLink accepts only strings, so I am also planning to import and use "getPathUrlAsString" but this need the "To" type from history, is this ok?

stories/LinkComponent-stories.tsx
gregaubert: Yes I think it will create a separate entry in storybook, but that's actually not so bad since it's a different variant that answer different use cases.

package.json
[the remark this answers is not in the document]
gregaubert: No we can still use 6.11.* in our dependencies, putting version ^6.0.0 in peerDependencies means we should be working fine with all the 6x versions.
If we are afraid it wouldn't work for 6.0 router we can keep 6.11.0.

src/components/links/LinkBase.tsx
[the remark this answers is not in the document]
gregaubert: But isSonarLink is only used in the external link which only support string. So you can pass it the `toAsString` that you have already computed.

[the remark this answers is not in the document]
revanshu-paliwal-sonarsource: Also, the eslint is complaining after this:
error  Using target="_blank" without rel="noreferrer" (which implies rel="noopener") is a security risk in older browsers: see https://mathiasbynens.github.io/rel-noopener/#recommendations  react/jsx-no-target-blank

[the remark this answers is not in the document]
gregaubert: Yeah, but I think it's fine to disable the rule for this case since we remove it only for links that we know

src/components/links/LinkBase.tsx
gregaubert: yes :)

src/components/links/LinkBase.tsx
gregaubert: Actually I just noticed we also have design tokens for that:

```
--echoes-text-decoration-none
--echoes-text-decoration-underline
```

package.json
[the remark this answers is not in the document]
revanshu-paliwal-sonarsource: I went with ^6.0.0, as mentioned by Greg above

stories/LinkComponent-stories.tsx
```diff
@@ -0,0 +1,100 @@
...
+export const LinkInsideParagraphWithStyle: Story = {
+  args: {
+    children: 'Link External inside paragraph',
+    to: 'https://abc.com/path/new',
+  },
+  render: (args) => (
+    <p style={{ color: 'grey', fontSize: '20px', fontWeight: '100' }}>
+      <Link {...args} />
+    </p>
+  ),
```
jeremy-davis-sonarsource, line 68: What I meant was surround it with text, like:
```
<p>
Hello, click <Link /> to go to the good place.
</p>
```

But your example already reveals an issue with the Icon alignment:
image
(or maybe it's the font size that triggers that?)

stories/LinkComponent-stories.tsx
```diff
@@ -0,0 +1,100 @@
...
+    children: 'Link External inside div',
+    to: 'https://abc.com/path/new',
+  },
+  render: (args) => (
+    <div style={{ display: 'flex', alignItems: 'center', justifyContent: 'center' }}>
+      <p>Paragraph 1</p>
+      <p>Paragraph 2</p>
+      <Link {...args} />
+      <span>Span 1</span>
+    </div>
```
jeremy-davis-sonarsource, line 98: How about adding a `gap`?
Also, this reveals some alignment issues (the link appears lower than the paragraphs)
It might be the default storybook paragraph styles, but we should try to simulate the products' context

stories/LinkSubduedComponent-stories.tsx
```diff
@@ -0,0 +1,62 @@
...
+import { LinkSubdued as LinkSubduedComp } from '@sonarsource/echoes-react';
+import type { Meta, StoryObj } from '@storybook/react';
+import { OpenNewTabIcon } from '../src/common/icons/OpenNewTabIcon';
+
+const meta: Meta<typeof LinkSubduedComp> = {
+  component: LinkSubduedComp,
+  title: 'Links',
+  args: {
+    to: '/path',
+  },
```
gregaubert, line 30: I would remove that since you provide the props everywhere anyway.

stories/LinkComponent-stories.tsx
```diff
@@ -0,0 +1,100 @@
...
+import { Link } from '@sonarsource/echoes-react';
+import type { Meta, StoryObj } from '@storybook/react';
+import { OpenNewTabIcon } from '../src/common/icons/OpenNewTabIcon';
+
+const meta: Meta<typeof Link> = {
+  component: Link,
+  title: 'Links',
+  args: {
+    to: '/path',
+  },
```
gregaubert, line 30: Same here

stories/LinkSubduedComponent-stories.tsx
```diff
@@ -0,0 +1,62 @@
...
+
+export const LinkSubduedWithIcon: Story = {
+  args: {
+    children: 'Link Subdued With Icon',
+    to: '/projects/new',
+    icon: <OpenNewTabIcon />,
+  },
+};
+
+export const LinkSubduedInsideParagraphWithStyle: Story = {
```
gregaubert, line 52: I would add some text before and after to show how it's inheriting the color and size

gregaubert (approved):
A few very minor things and a question for Marcio but otherwise it looks good! Well done! 😄

stories/LinkComponent-stories.tsx
gregaubert: I think we shouldn't care about that for now.
Our products still only support single-size icon of 16px only.

We should anyway revisit a bit the implementation of this when we implement the new icons with the 3 sizes and the typography components that should be used for sizing things.

Maybe just reduce the font-size back to 1rem on this story for now and use a different font-decoration instead like italic.

gregaubert: By the way, I created DS-202 for that.

src/components/links/__tests__/Link-test.tsx
```diff
@@ -0,0 +1,164 @@
...
+  'react-intl',
+  () =>
+    ({
+      ...jest.requireActual('react-intl'),
+      useIntl: () => ({
+        formatMessage: ({ id }: { id: string }, values = {}) =>
+          [id, ...Object.values(values)].join('.'),
+      }),
+    }) as IntlShape,
+);
```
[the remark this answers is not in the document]
revanshu-paliwal-sonarsource: If I remove this the test are failing with error:
 Error: Uncaught [Error: [React Intl] Could not find required `intl` object. <IntlProvider> needs to exist in the component ancestry.]

stories/LinkComponent-stories.tsx
revanshu-paliwal-sonarsource: Lets discuss this tomorrow, I did not completely understood this one

src/components/links/__tests__/Link-test.tsx
```diff
@@ -19,24 +19,26 @@
...
```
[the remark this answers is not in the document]
revanshu-paliwal-sonarsource: Woah for a second I got confused with "traductions" (translation) 😅, just confirming do you mean something like below:
<MemoryRouter initialEntries={initialEntries}>
      <Routes>
        <Route
          element={
            <IntlProvider locale="en">
              {component}
              <ShowPath />
            </IntlProvider>
          }
          path="/initial"
        />
        <Route element={<ShowPath />} path="/second" />
      </Routes>
    </MemoryRouter>,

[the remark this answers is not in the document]
revanshu-paliwal-sonarsource: And yes it works 👍

src/components/links/__tests__/Link-test.tsx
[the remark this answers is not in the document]
revanshu-paliwal-sonarsource: I added a IntlProvider in custom renderer and it worked

src/components/links/__tests__/Link-test.tsx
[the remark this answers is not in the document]
gregaubert: Ahah sorry for the frenglish! Nice if it work!

src/components/links/__tests__/Link-test.tsx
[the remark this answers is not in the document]
jeremy-davis-sonarsource: I wasn't clear in my comment. but that's what I meant

jeremy-davis-sonarsource (approved):

PR #31, opened 2023-12-13
DS-191 define release process

lucas-paulger-sonarsource:


docs/RELEASING.md
```diff
@@ -0,0 +1,5 @@
+## Release Process
+
+- [Create a Jira release ticket](https://jira.sonarsource.com/projects/DS) like [this one](https://sonarsource.atlassian.net/browse/DS-200)
```
lucas-paulger-sonarsource, line 3: obviously, we haven't created an example yet... but I think it would be valuable. Any ideas for alternatives?

docs/RELEASING.md
```diff
@@ -0,0 +1,5 @@
...
+- [Publish a new GitHub release](https://github.com/SonarSource/echoes-react/releases/new)
+- [Bump the package version for the next development iteration](https://github.com/SonarSource/echoes-react/blob/main/package.json#L3)
```
lucas-paulger-sonarsource, line 5: an example here will be useful too

docs/RELEASING.md
jeremy-davis-sonarsource: To avoid confusion, I'd make it an obvious placeholder. 
I think an example is the best, but maybe we can list the fields to be filled with a short explanation?

jeremy-davis-sonarsource (changes requested):
What about adding a link to it from the README ?

jeremy-davis-sonarsource (approved):
Nice!

PR #32, opened 2023-12-13
DS-193 yarn zero-install

lucas-paulger-sonarsource:


lucas-paulger-sonarsource:
I had to rebuild the cached docker image but the build is fast when no deps are changed
https://cirrus-ci.com/build/5016209096179712

but we still need to use a cache for `.yarn/unplugged` since it is created at install and we apparently need it, but it should be ignored from git
https://github.com/yarnpkg/berry/issues/454

I also wanted to look for examples of this in the wild, of course yarn is using this approach
https://github.com/yarnpkg/berry/tree/master/.yarn/cache

and some others
https://github.com/cosmos/cosmjs/tree/main/.yarn
https://github.com/EventStore/Design-System/tree/main/.yarn/cache

.gitignore
```diff
@@ -7,8 +7,8 @@
 
 # Swap the comments on the following lines if you wish to use zero-installs
 # Documentation here: https://yarnpkg.com/features/zero-installs
```
jeremy-davis-sonarsource, line 9: We should fix the comment here, it's confusing now

jeremy-davis-sonarsource (approved):
Seems to be the way forward

gregaubert (approved):
Looks good!

PR #36, opened 2023-12-15
DS-188 DS-203 Add Checkbox and Spinner components

gregaubert:


src/components/checkbox/CheckboxIcon.tsx
revanshu-paliwal-sonarsource, line 1: Should this be part of icons?

src/common/helpers/styles.ts
```diff
@@ -0,0 +1,32 @@
...
+  position: absolute;
+  width: 1px;
+  height: 1px;
+  padding: 0;
+  margin: -1px;
+  overflow: hidden;
+  clip: rect(0, 0, 0, 0);
+  white-space: nowrap;
+  border-width: 0;
+`;
```
jeremy-davis-sonarsource, line 32: NP: How about `screenReaderOnly`? It took me a while to figure out what it meant

src/components/spinner/Spinner.tsx
```diff
@@ -0,0 +1,126 @@
...
+      id: 'loading',
+      defaultMessage: 'Loading...',
+      description: 'aria-label text, to indicate that there is spinner rotating in this place',
+    }),
+    isLoading = true,
+  } = props;
+
+  if (customSpinner) {
+    return <>{isLoading ? customSpinner : children}</>;
+  }
```
jeremy-davis-sonarsource, line 52: Do we have a legitimate use case for a custom spinner? 
In SQ we use the prop once, and it's to wrap the spinner in a `tr` + `td`, which we could probably do differently (maybe a `wrapper` prop?).
In that specific case, we're also losing the a11y features :unamused:

stories/Spinner-stories.tsx
```diff
@@ -0,0 +1,63 @@
...
+
+export const WithPlaceHolder: Story = {
+  args: {
+    isLoading: false,
+    hasPlaceholder: true,
+  },
+  render: (args) => (
+    <p>
+      There is a Spinner placeholder here:
+      <Spinner {...args} />
```
jeremy-davis-sonarsource, line 59: NP: Add margins to make it look nicer :nail_care:

src/components/spinner/__tests__/Spinner-test.tsx
revanshu-paliwal-sonarsource, line 1: We need to add jest-axe test for spinner

src/components/checkbox/__tests__/Checkbox-test.tsx
revanshu-paliwal-sonarsource, line 1: We need to add jest-axe test for checkbox

src/components/checkbox/Checkbox.tsx
```diff
@@ -0,0 +1,153 @@
...
+ * Inc., 51 Franklin Street, Fifth Floor, Boston, MA  02110-1301, USA.
+ */
+import styled from '@emotion/styled';
+import * as RadixCheckbox from '@radix-ui/react-checkbox';
+import { useCallback } from 'react';
+import { Spinner } from '../spinner';
+import { CheckboxIcon } from './CheckboxIcon';
+
+interface Props {
+  checked: boolean | 'indeterminate';
```
revanshu-paliwal-sonarsource, line 27:
```suggestion
  isChecked: boolean | 'indeterminate';
```

src/components/checkbox/Checkbox.tsx
```diff
@@ -0,0 +1,153 @@
...
+}
+
+export function Checkbox(props: Readonly<Props>) {
+  const { checked, className, label, ariaLabel, title, id, isDisabled, isLoading = false } = props;
+  const { onCheck, onFocus } = props;
+
+  const handleChange = useCallback(
+    (checked: boolean | 'indeterminate') => {
+      if (!isDisabled && !isLoading) {
+        onCheck(checked, id);
```
revanshu-paliwal-sonarsource, line 49: In SQ this line is onCheck(!checked, id);
Need to check why is the difference

src/components/checkbox/Checkbox.tsx
```diff
@@ -0,0 +1,153 @@
...
+            <CheckboxIcon checked={checked} />
+          </CheckboxIndicator>
+        </CheckboxRoot>
+      </Spinner>
+      {label && <Label>{label}</Label>}
+    </CheckboxContainer>
+  );
+}
+
+const CheckboxContainer = styled.span`
```
revanshu-paliwal-sonarsource, line 79: NIT: In SQ we are also using "user-select: none", I feel its not needed but just putting it here

src/components/checkbox/Checkbox.tsx
```diff
@@ -0,0 +1,153 @@
...
+  display: inline-flex;
+  align-items: center;
+  vertical-align: text-bottom;
+
+  &[aria-disabled='true'] {
+    color: var(--echoes-color-text-disabled);
+    pointer-events: none;
+  }
```
revanshu-paliwal-sonarsource, line 87: We should also add cursor on hover

revanshu-paliwal-sonarsource (commented):
Minor change, I am also trying to check few cases of SQ which is taking time, will add comment later if I find anything

src/components/checkbox/CheckboxIcon.tsx
gregaubert: I don't think so, because it's "smart icon" that chooses what to render based on a prop. It's more of an "icon selector".
Icons that are part of the icons folder should be plain dumb and just render something.

src/components/spinner/__tests__/Spinner-test.tsx
gregaubert: Ah yes! good point!

src/components/checkbox/Checkbox.tsx
gregaubert: I'm not sure about this one, since it isn't a pure boolean with the 'indeterminate' state...
What's your opinion there @jeremy-davis-sonarsource ?

src/components/checkbox/Checkbox.tsx
gregaubert: It's different because the radix-ui implementation already send us the new state, if we had a checked=true, in the handle change we are gonna receive a checked=false, and vice-versa.

So the feature remains the same.

src/components/checkbox/Checkbox.tsx
gregaubert: Yeah same in SC and I'm not sure it really makes sense to prevent selecting the label next to the checkbox. I would keep it like this.

src/components/checkbox/Checkbox.tsx
gregaubert: I was wondering about that since the radix-ui implem and examples didn't have the hand cursor.

But [reading a bit about it](https://medium.com/simple-human/buttons-shouldnt-have-a-hand-cursor-b11e99ca374b) it seems like the hand cursor is actually for links only and doesn't mean "interactive". So I would keep it like this.

gregaubert: And maybe we should also review our other usages of the cursor:pointer by the way.

src/components/spinner/Spinner.tsx
gregaubert: Yeah I think that's a good point, I just checked in SC and it's the same we need a wrapper instead. The other cases can be easily rewritten.

src/components/checkbox/Checkbox.tsx
jeremy-davis-sonarsource: Yeah, it's not a boolean. `checked` is fine

src/components/spinner/Spinner.tsx
```diff
@@ -0,0 +1,119 @@
...
+import { keyframes } from '@emotion/react';
+import styled from '@emotion/styled';
+import { ReactNode } from 'react';
+import { FormattedMessage } from 'react-intl';
+import { screenReaderOnly } from '~common/helpers/styles';
+
+interface Props {
+  className?: string;
+  children?: ReactNode;
+  label?: ReactNode;
```
revanshu-paliwal-sonarsource, line 30: NIT: label is a react node but we are using it as a boolean. Also, I feel label prop for a spinner might be confusing?

stories/Spinner-stories.tsx
```diff
@@ -0,0 +1,55 @@
...
+export default meta;
+
+type Story = StoryObj<typeof meta>;
+
+export const Basic: Story = {
+  args: {
+    isLoading: true,
+    children: <span>isLcomplete</span>,
+  },
+};
```
revanshu-paliwal-sonarsource, line 41: Should we add a story for label usecase too?

gregaubert: Good catch! I just added it! ;)

jeremy-davis-sonarsource (approved):
Awesome!

PR #38, opened 2023-12-15
Cleanup

jeremy-davis-sonarsource:


.yarnrc.yml
```diff
@@ -14,3 +14,14 @@ progressBarStyle: simba
 tsEnableAutoTypes: true
 
 yarnPath: .yarn/releases/yarn-4.0.2.cjs
+
+supportedArchitectures:
+  os:
+    - "current"
```
gregaubert, line 20: I wonder what's the impact of "current" here...

gregaubert (approved):

.yarnrc.yml
jeremy-davis-sonarsource: I assume it detects the system you are on and fetches the appropriate package.
If someone uses windows, they'll be able to run `yarn`, which will install the additional packages.
That way we can add to the list as people join the project

PR #40, opened 2024-01-26
DS-243 Update usage readme for jest

gregaubert:


gregaubert:
@jeremy-davis-sonarsource In the end the builds with CJS and UMD don't work at all when trying to use them with jest in the product. I don't know why though. So I updated the doc a bit and just changed the package.json to use `imports` instead of `main`, `imports` is the more recent way of doing it, it's better because it's also limiting usage to only what you export, users won't be able to do things like `@sonarsource/echoes-react/components/checkbox` (unless we explicitly allow it in the `imports` definition.

README.md
```diff
@@ -2,6 +2,32 @@
...
+## Installation
+
+```bash
+yarn add @sonarsource/echoes-react
+```
+
+## Usage
+
+Make sure to setup the `IntlProvider` from `react-intl` at the top of your app. See [this page](https://formatjs.io/docs/react-intl/components/#intlprovider) for more information.
+
```
jeremy-davis-sonarsource, line 14: NP: "root" rather than "top"?

README.md
```diff
@@ -2,6 +2,32 @@
...
+```ts
+import { Checkbox } from '@sonarsource/echoes-react';
+```
+
+See available components and usage in storybook: https://echoes-react.netlify.app/
+
+### Make it work in Jest
+
+The lib only provide es module bundle, so if you use Jest for your tests make sure your transform preprocessor goes through `echoes-react` to make it understood by Node.js, you can do that by adding an exception in your `transformIgnorePatterns`, for example:
```
jeremy-davis-sonarsource, line 25:
```suggestion
The lib only provides es module bundle. If you use Jest for your tests (or a similar library) make sure your transform preprocessor goes through `echoes-react` to make it runnable on Node.js. 
You can do that by adding an exception in your `transformIgnorePatterns`, for example:
```

jeremy-davis-sonarsource (approved):
LGTM!

PR #42, opened 2024-02-20
DS-186 Update Link components with latest figma specs

gregaubert:


src/components/links/LinkStandalone.tsx
```diff
@@ -18,13 +18,43 @@
...
 import styled from '@emotion/styled';
+import { LinkProps } from '.';
 import { LinkBaseStyled } from './LinkBaseStyled';
 
-export const LinkSubdued = styled(LinkBaseStyled)`
-  --color: 'currentColor';
-  --hover: var(--echoes-color-background-accent-default-hover);
-  --active: var(--echoes-color-background-accent-default-active);
+interface Props extends LinkProps {
+  iconLeft?: React.ReactNode;
```
david-cho-lerat-sonarsource, line 27: why not `leftIcon`?

stories/LinkStandalone-stories.tsx
```diff
@@ -18,42 +18,40 @@
...
+    children: 'Standalone Link',
+    highlight: LinkHighlight.Accent,
     to: '/projects/new',
   },
 };
 
-export const LinkSubduedWithIcon: Story = {
+export const LinkStandaloneWithIcon: Story = {
   args: {
     children: 'Link Subdued With Icon',
```
david-cho-lerat-sonarsource, line 52: shouldn't this say 'Link Standalone With Icon'?

src/components/links/Link.tsx
```diff
@@ -18,15 +18,8 @@
...
-export const Link = styled(LinkBaseStyled)`
-  font-weight: var(--echoes-font-weight-semi-bold);
-
-  --color: var(--echoes-color-background-accent-default);
-  --hover: var(--echoes-color-background-accent-default-hover);
-  --active: var(--echoes-color-background-accent-default-active);
-`;
+export const Link = LinkBaseStyled;
 
 Link.displayName = 'Link';
```
david-cho-lerat-sonarsource, line 25: this file doesn't do much now - do we still need the two?

src/components/links/LinkStandalone.tsx
gregaubert: Not a particular reason, it was already named like that in figma, and in case we add a iconRight later both of them will be next to each other in autocompletion, which I think is nice.

src/components/links/Link.tsx
gregaubert: Yeah I wasn't sure.
I could also export it directly from `LinkBaseStyled.tsx` but it didn't seem very clean because we wouldn't have a file per variant anymore.

I could also rename `LinkBaseStyled.tsx` to `Link.tsx` and export from there, but I kind of liked that we could identify quickly what's the base style for the links...

It seems fine to me to have this almost empty file, it is pretty explicit that currently, the default link doesn't override anything from the base. But I'm ok to change it to something else if you want?

david-cho-lerat-sonarsource (approved):
LGTM :+1:

PR #46, opened 2024-02-21
DS-188 Update Checkbox component with latest figma changes

gregaubert:


README.md
```diff
@@ -33,14 +33,17 @@ transformIgnorePatterns: [`/node_modules/(?!@sonarsource/echoes-react)`],
...
+We recommend VSCode to work on this project.
+There is a `.vscode` folder containing:
 
 - A list of recommended extensions
   - Install them through the marketplace
 - A template for necessary (and recommended) settings: `settings.template.json`
   - Copy it and save it as `settings.json`
   - You must open the project directly. Adding its folder in an existing workspace might not work, as the typescript configuration must be defined at workspace level.
 
+You should also [setup your vscode to work](https://yarnpkg.com/getting-started/editor-sdks) with the `yarn` pnp setup, using the following command: `yarn dlx @yarnpkg/sdks vscode`
```
david-cho-lerat-sonarsource, line 45: setup is a noun: 'You should also set up'

src/components/checkbox/Checkbox.tsx
```diff
@@ -28,19 +28,42 @@ interface Props {
...
+}
+
+interface PropsWithoutLabel extends Props {
+  label?: never;
+  ariaLabel: string;
+}
+
+export function Checkbox(props: Readonly<PropsWithLabel | PropsWithoutLabel>) {
+  const {
+    checked,
```
david-cho-lerat-sonarsource, line 56: NP: we don't have a rule to sort alphabetically? I find it easier to parse

src/components/checkbox/Checkbox.tsx
```diff
@@ -65,32 +88,35 @@ export function Checkbox(props: Readonly<Props>) {
           id={id}
           onCheckedChange={handleChange}
           onFocus={onFocus}
-          title={title}>
+          title={title}
+          {...(hasError && checked === false ? { 'data-error': true } : {})}>
```
david-cho-lerat-sonarsource, line 92: if it's in error, you can still check it, though - does that make sense?

src/components/checkbox/Checkbox.tsx
```diff
@@ -101,45 +127,49 @@ const CheckboxRoot = styled(RadixCheckbox.Root)`
...
-      background-color: var(--echoes-color-background-accent-default-disabled);
+      background-color: var(--echoes-color-background-default-disabled);
+      border-color: var(--echoes-color-background-default-disabled);
     }
 
     &:focus,
     &:focus-visible {
-      outline: var(--echoes-color-border-disabled) solid 2px;
-      outline-offset: var(--echoes-dimension-space-25);
+      outline: var(--echoes-color-border-disabled) solid var(--echoes-focus-border-width-default);
```
david-cho-lerat-sonarsource, line 143: didn't @marciopmoreira6 say that the focus should always look the same?

marciopmoreira6 (commented):
Looks good to me in Storybook. Very clean. 
Well done 👍

david-cho-lerat-sonarsource (changes requested):
Just a few details/questions in the comments.
Also: when in 'indeterminate' state, shouldn't a click uncheck rather than check?
For me 'indeterminate' is a special kind of 'checked' - WDYT?

david-cho-lerat-sonarsource:
Also: the vertical alignment seems different from the Figma
image
image

src/components/checkbox/Checkbox.tsx
gregaubert: Following the conversation with Marcio in Figma, the use case for error is mainly when you must check it for license purposes or that kind of thing.
So it doesn't seem to make sense to allow an error state for checked checkbox, hence the condition.
I'll add comment.

src/components/checkbox/Checkbox.tsx
gregaubert: Indeed, I'll fix it!

david-cho-lerat-sonarsource (approved):
LGTM :+1:

PR #49, opened 2024-02-21
DS-172 Update radio button component with the latest Figma changes

david-cho-lerat-sonarsource:


gregaubert (approved):
I think we are good! Just one minor thing!

PR #51, opened 2024-02-23
DS-186 Fix LinkStandalone focus style

gregaubert:
Focus style was wrong for the standalone link.
And the weight was wrong for all the links except the accent, they should all be semi-bold.


david-cho-lerat-sonarsource (approved):
LGTM :+1:

PR #54, opened 2024-02-26
DS-186 Update Link components with better defaults

gregaubert:


src/components/links/LinkBase.tsx
```diff
@@ -31,6 +31,7 @@ export enum LinkHighlight {
   Accent = 'accent',
   Default = 'default',
   Subdued = 'subdued',
+  CurrentColor = 'current-color',
```
david-cho-lerat-sonarsource, line 34: my OCD would prefer this between Accent and Default :grin:

PR #55, opened 2024-02-26
DS-203 Add className wrapper to Spinner API

gregaubert:


src/components/spinner/Spinner.tsx
```diff
@@ -20,26 +20,34 @@
...
 import { keyframes } from '@emotion/react';
 import styled from '@emotion/styled';
+import classNames from 'classnames';
 import { ReactNode } from 'react';
 import { FormattedMessage } from 'react-intl';
 import { screenReaderOnly } from '~common/helpers/styles';
 
 interface Props {
   className?: string;
+  classNameWrapper?: string;
```
david-cho-lerat-sonarsource, line 30: using `wrapperClassName` would be less confusing and more accurate IMO

david-cho-lerat-sonarsource (changes requested):
This isn't really adding a "large spinner", but rather a "stylable spinner", right? You can use that custom style to make it large, or anything else

gregaubert:
> This isn't really adding a "large spinner", but rather a "stylable spinner", right? You can use that custom style to make it large, or anything else

Yeah indeed, at first I added a isLarge prop and then I noticed there were other use cases and changed the implem and forgot to rename the PR.

src/components/spinner/Spinner.tsx
gregaubert: Ok I'll change it!

david-cho-lerat-sonarsource (approved):
LGTM, thanks! :+1:

PR #60, opened 2024-03-18
DS-196 Generate expected i18n keys list

jeremy-davis-sonarsource:


package.json
```diff
@@ -23,7 +23,7 @@
     "build": "vite build",
     "build-tokens": "node design-tokens/build.js",
     "build-storybook": "vite build & storybook build",
-    "intl": "formatjs extract src/components/**/*.ts*",
+    "intl-keys": "formatjs extract src/components/**/*.ts* --out-file i18n/keys.json",
```
gregaubert, line 26: Should we make it part of the build command?

gregaubert (approved):
Nice! :) I actually didn't know we had a command already.
I guess it would also make sense to document a link to that file in the readme where we talk about the react-intl provider.

package.json
gregaubert: Can be another ticket then

jeremy-davis-sonarsource: I considered it, and I think the `releasability check` is better

PR #63, opened 2024-03-19
DS-270 Maintain Checkbox label position between loading and non-loading states

david-cho-lerat-sonarsource:


jeremy-davis-sonarsource (approved):
LGTM!

src/components/checkbox/Checkbox.tsx
```diff
@@ -17,28 +17,27 @@
...
-
-  hasError?: boolean;
+  innerClassName?: string;
   isDisabled?: boolean;
   isLoading?: boolean;
-
+  label?: string;
   onCheck: (checked: boolean | 'indeterminate', id?: string) => void;
   onFocus?: VoidFunction;
+  title?: string;
```
gregaubert, line 40: @jeremy-davis-sonarsource @david-cho-lerat-sonarsource should we enable an eslint rule for that?

david-cho-lerat-sonarsource: I wouldn't mind always having the props in alpha order, at least it's a system. I know the opposing arguments of grouping them "semantically", or "functions together", or "important first", etc, but in real life when no order is enforced none of this is followed in the long run and we just end up with "whatever". I'd much prefer to enforce the alpha order and be done with it.

jeremy-davis-sonarsource: Sarath (?) mentioned having the same VScode config for everyone (and every project!), which think makes sense, and will solve this aspect as well

gregaubert: I think this is enforced by an eslint rule on SC, not sure the vscode config will help there.
But I guess we also want a common eslint config for all projects.

PR #68, opened 2024-03-21
DS-288 Add the custom icons fonts

gregaubert:


src/components/icons/IconWrapper.tsx
gregaubert, line 1: Not sure about the pattern here for supporting two different font-families, if you have better ideas feel free to suggest

src/components/icons/IconWrapper.tsx
```diff
@@ -47,3 +47,21 @@ const StyledSpan = styled.span`
...
+  font-family: 'Echoes Icons';
+  font-size: 1.2em;
+  font-weight: normal;
+  font-style: normal;
+  line-height: 1;
+  letter-spacing: normal;
+  text-transform: none;
+  white-space: nowrap;
+  word-wrap: normal;
+  direction: ltr;
```
david-cho-lerat-sonarsource, line 62: do we really need this one (`direction`) ?

gregaubert: Probably not, this whole style comes from the class provided by the material UI lib, there are probably many things we can clean up.

david-cho-lerat-sonarsource (approved):
Cool. Maybe we can also display IconCloud in Storybook?

src/components/icons/IconWrapper.tsx
jeremy-davis-sonarsource: I don't see any better pattern

gregaubert:
> Cool. Maybe we can also display IconCloud in Storybook?

We can add once we merge with the Story from Jay, I don't have it yet on this PR

jeremy-davis-sonarsource (approved):
LGTM!

PR #70, opened 2024-03-22
DS-289 Update node modules

gregaubert:


package.json
```diff
@@ -43,62 +43,63 @@
...
+    "@testing-library/dom": "9.3.4",
     "@testing-library/jest-dom": "6.1.4",
     "@testing-library/react": "12.1.5",
-    "@testing-library/user-event": "14.5.1",
-    "@tokens-studio/sd-transforms": "0.11.10",
-    "@types/eslint": "8.44.7",
-    "@types/jest": "29.5.10",
-    "@types/jest-axe": "3.5.8",
+    "@testing-library/user-event": "14.5.2",
+    "@tokens-studio/sd-transforms": "0.14.4",
```
gregaubert, line 66: The last version is 0.15.0 but it relies on an unreleased version of style-dictionary. I figured it would be safer to wait for a style-dictionary to do an official release before updating.

package.json
```diff
@@ -43,62 +43,63 @@
...
-    "typescript": "5.2.2",
-    "vite": "5.0.0",
-    "vite-plugin-dts": "3.6.3"
+    "react-intl": "6.6.2",
+    "react-router-dom": "6.22.3",
+    "rollup": "4.13.0",
+    "rollup-plugin-license": "3.3.1",
+    "storybook": "8.0.4",
+    "style-dictionary": "3.9.2",
+    "typescript": "5.3.3",
```
gregaubert, line 100: The last version is 5.4.3, but it's generating tons of errors on every file like if the whole react typing was missing. It's probably a conflict or incompatibility, I tried to resolve it while keeping react 17 types, but didn't manage so far.

jeremy-davis-sonarsource (approved):
Thanks for doing this!

package.json
jeremy-davis-sonarsource: I'm not surprised

gregaubert: Why? Is there something specific about this version?

PR #73, opened 2024-04-02
DS-292 Test custom font integration

david-cho-lerat-sonarsource:


src/components/icons/echoes-icons.woff2
gregaubert, line 1: Can we re-export the font with a proper name? Right now it's named "asdasd" when installed on the system:
image

gregaubert (approved):
We would just need a cleaner font name but for the POC it's fine.

gregaubert:
Well, since we are not in a POC anymore, we should have the proper font name before merging

gregaubert (approved):
👍🏻

PR #75, opened 2024-04-03
DS-198 Improve generated files validation scripts

gregaubert:


jeremy-davis-sonarsource (changes requested):
The error message is quite cryptic if you don't already know about it.
Also, if at some point it runs in parallel to some other job, it'll be even more confusing.

A simple way could be to have a script that echoes the error and call it like:

```
yarn build-intl-keys && git diff --exit-code --compact-summary i18n/keys.json || node scripts/printErrorMessage.js
```

jeremy-davis-sonarsource:
Also, we could use the `--quiet` flag instead of `--exit-code --compact-summary`

gregaubert:
> Also, we could use the `--quiet` flag instead of `--exit-code --compact-summary`

`--quiet` means no output at all, from what I could test

I'll create two scripts that write a nice error message instead.

gregaubert:
Improved error messages:
image
image

config/scripts/validate-intl-keys.sh
```diff
@@ -0,0 +1,11 @@
+#!/bin/bash
+
+yarn build-intl-keys
+
+if (git diff --quiet i18n/keys.json > /dev/null); then
+  echo "i18n/keys.json is up to date."
```
david-cho-lerat-sonarsource, line 6: isn't this then/else inverted? If we get into the `then` it should mean we're out of date, right?

config/scripts/validate-tokens.sh
```diff
@@ -0,0 +1,11 @@
+#!/bin/bash
+
+yarn build-tokens > /dev/null
+
+if (git diff --quiet src/generated > /dev/null); then
+  echo "Design-tokens are up to date."
```
david-cho-lerat-sonarsource, line 6: isn't this then/else inverted? If there's no diff then the exit code of `git diff` is `0` - shouldn't that trigger the `else` rather than the `then`?

david-cho-lerat-sonarsource (changes requested):
I believe the logic may be inverted in the two check scripts

config/scripts/validate-intl-keys.sh
jeremy-davis-sonarsource: it works for me

jeremy-davis-sonarsource (approved):
Awesome work!

config/scripts/validate-tokens.sh
gregaubert: `git diff --quiet` exit code is 0 when there is no diff with means it enters the first if close (no error)
And the exit code is 1, which means error/failure, and that's assimilated to false to it enters the else.

gregaubert: it's not the same as returning a boolean true or false or a number, it's an exit code, the exit code value doesn't really matter in this case. You could read it as `if no error in git diff`.

config/scripts/validate-intl-keys.sh
gregaubert: yes it's works, it's just bash exit code that's misleading in the condition

config/scripts/validate-tokens.sh
david-cho-lerat-sonarsource: OK, that's confusing :/ - thanks for the explanation!

david-cho-lerat-sonarsource (approved):

PR #77, opened 2024-04-04
DS-280 Implement 2nd batch of icons

david-cho-lerat-sonarsource:


src/components/icons/IconStar.tsx
```diff
@@ -22,9 +22,5 @@ import { IconMaterialWrapper, IconProps } from './IconWrapper';
 
 export function IconStar(props: Readonly<IconProps>) {
   // This is Material Symbols' "kid_star" icon
-  return (
```
gregaubert, line 25: There are actually two versions of star

gregaubert (approved):
Nice!

src/components/icons/IconStar.tsx
david-cho-lerat-sonarsource: now handled with the export of `isFilled`, same for `IconHome`

gregaubert (approved):
Nice update!

PR #78, opened 2024-04-05
DS-278 Update icons naming, cleanup old icons and add tests

gregaubert:


src/components/icons/IconWrapper.tsx
```diff
@@ -19,24 +19,46 @@
...
+    children,
+    isFilled, // extracted here to prevent it from leaking to the span element
+    ...rest
+  } = props;
 
   return (
-    <span className={className} {...rest}>
+    <span
+      // 'aria-label' is not a valid prop, but it might be used by the consumer when wrapping the component with emotion
+      aria-hidden={isStringDefined(ariaLabel ?? nativeAriaLabel) ? 'false' : 'true'}
```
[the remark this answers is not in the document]
gregaubert: Drop this support

src/components/icons/IconWrapper.tsx
```diff
@@ -19,24 +19,46 @@
...
-  const { children, className, ...rest } = props;
+type IconWrapperProps = JSX.IntrinsicElements['span'];
+
+function IconBase(props: Readonly<IconWrapperProps & IconFilledProps>) {
+  const {
+    ariaLabel,
+    'aria-label': nativeAriaLabel,
+    children,
+    isFilled, // extracted here to prevent it from leaking to the span element
+    ...rest
```
gregaubert, line 46: Make className explicit

gregaubert: Drop `...rest`

src/components/icons/IconWrapper.tsx
```diff
@@ -19,24 +19,46 @@
...
+  */
+}
+
+export interface IconFilledProps extends IconProps {
+  isFilled?: boolean;
+}
 
-function IconBase(props: Readonly<IconProps & { children: React.ReactNode }>) {
-  const { children, className, ...rest } = props;
+type IconWrapperProps = JSX.IntrinsicElements['span'];
```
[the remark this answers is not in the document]
gregaubert: Change that to a more restrictive props list

src/components/icons/IconWrapper.tsx
```diff
@@ -19,24 +19,46 @@
...
+import { isStringDefined } from '~common/helpers/types';
 
-export type IconProps = Omit<JSX.IntrinsicElements['span'], 'children'>;
+export interface IconProps {
+  ariaLabel?: string;
+  className?: string;
+  /* To discuss?
+  ['data-guiding-id']?: string;
+  fill?: ThemeColors | CSSColor;
+  style?: React.CSSProperties;
```
gregaubert, line 30: color prop that accepts only color design tokens.

src/components/icons/__tests__/IconWrapper-test.tsx
```diff
@@ -0,0 +1,77 @@
...
+  const icon = screen.getByRole('img', { hidden: true });
+  expect(icon).toContainHTML('&#xE3A6;');
+  expect(icon).toHaveStyle({
+    'font-family': `'Material Symbols Rounded'`,
+  });
+  expect(icon).not.toHaveStyle({
+    'font-variation-settings': `'FILL' 1`,
+    'font-family': `'Echoes'`,
+  });
+  expect(container).toMatchSnapshot();
```
david-cho-lerat-sonarsource, line 38: NP: do we really want to have snapshots? Are they that useful?

stories/Icons-stories.tsx
david-cho-lerat-sonarsource, line 1: NP for line 42: `{['IconDot', 'IconHome', 'IconStar'].includes((Icon as React.FC).displayName ?? '') && (`

david-cho-lerat-sonarsource: maybe we should detect those that accept `IconFilledProps` somehow instead

david-cho-lerat-sonarsource (approved):
Looking great - nice job :+1:

src/components/icons/__tests__/IconWrapper-test.tsx
gregaubert: I wanted them here to check on the calculations in the CSS part, seems pretty heavy to do it with expects.

gregaubert: I see now that we can easily drop one of them, only one is useful IMO

src/components/links/LinkBase.tsx
```diff
@@ -22,8 +22,8 @@ import styled from '@emotion/styled';
 import React, { HTMLAttributeAnchorTarget } from 'react';
 import { useIntl } from 'react-intl';
 import { Link as RouterLink, LinkProps as RouterLinkProps } from 'react-router-dom';
-import { isSonarLink } from '../../common/helpers/url';
-import { OpenNewTabIcon } from '../../common/icons/OpenNewTabIcon';
+import { isSonarLink } from '~common/helpers/url';
+import { IconLinkExternal } from '../icons';
```
jeremy-davis-sonarsource, line 26: Here it's okay (icons are low-level), but it's better to import straight from the module rather than from an index:
otherwise, it can lead to circular dependencies

package.json
```diff
@@ -105,7 +105,6 @@
     "vite-plugin-dts": "3.8.1"
   },
   "dependencies": {
-    "@primer/octicons-react": "19.8.0",
```
jeremy-davis-sonarsource, line 108: Woohoo :axe:

src/common/helpers/design-tokens.ts
```diff
@@ -0,0 +1,29 @@
...
+
+import designTokensColors from '~generated/design-tokens-colors.json';
+
+export type DesignTokens = {
+  [K in keyof typeof designTokensColors]: `--${K}`;
+}[keyof typeof designTokensColors];
+
+export type DesignTokensColors = {
+  [K in DesignTokens]: K extends `--echoes-color-${string}` ? K : never;
+}[DesignTokens];
```
jeremy-davis-sonarsource, line 29: For readability, I'd prefer not to have the `--` prefix in the values.
We could instead prefix it in `IconWrapper` :
```
${({ color }) => isDefined(color) && `color: var(--${color});`}
```

WDYT?

jeremy-davis-sonarsource (approved):
Great work! A couple of suggestions only

src/common/helpers/design-tokens.ts
gregaubert: We can easily remove it from the value, I was thinking we want all usage of the design token to look the same.
But maybe it's already not the case even with that since in CSS there is the `var()` also...
Alright I'll remove the `--` from he type.

PR #80, opened 2024-04-11
DS-302 Rename IconFoldDown and IconFoldUp to IconUnfoldDown and IconUnfoldUp

david-cho-lerat-sonarsource:


gregaubert (approved):
Can you target master, merge it on master and then cherry-pick it to branch-0.2 ?
I think that should be the way of working with those branches. We basically always work on master branch and backport stuff that we want to create a patch version for.

PR #86, opened 2024-04-23
DS-336 Add layer 3 overlay max width design token

marciopmoreira6:


design-tokens/tokens/layer3/base.json
```diff
@@ -1 +1,14 @@
-{}
\ No newline at end of file
+{
+  "echoes": {
+    "sizes": {
+      "overlays": {
+        "maxWidth": {
+          "default": {
+            "value": "{echoes.dimension.space.800}",
```
gregaubert, line 7: @marciopmoreira6 it should reference `size.5000` right?
I'll fix it with a commit, I need to fix the build for layer 3 anyway.

design-tokens/build.js
```diff
@@ -36,11 +36,14 @@ const licenseHeader = fs.readFileSync(`config/license/LICENSE-HEADER.txt`, 'utf-
 const designTokenGroups = JSON.parse(
   fs.readFileSync(`${DESIGN_TOKENS_PATH}/$themes.json`, 'utf-8'),
 );
+const baseDesignTokenGroup = designTokenGroups.filter(
+  ({ group, name }) => group === 'Sonar' && name === 'base',
+)[0];
```
jeremy-davis-sonarsource, line 41: NP:
```
const baseDesignTokenGroup = designTokenGroups.find(
  ({ group, name }) => group === 'Sonar' && name === 'base',
);
```

And maybe we should exit if we don't find it?

jeremy-davis-sonarsource (changes requested):

design-tokens/tokens/layer3/base.json
marciopmoreira6: thanks

design-tokens/build.js
gregaubert: Good catch, I'll update it.

> And maybe we should exit if we don't find it?

I don't think it's worth going through the hassle of handling these cases, it shouldn't happen, if it does, it's not gonna generate the expected files and/or break and that's fine.

jeremy-davis-sonarsource (approved):

PR #88, opened 2024-04-25
DS-360 Implement font subsetting to reduce size of material-symbols font to only what we need

gregaubert:
Before:
image

After:

image


.storybook/main.ts
```diff
@@ -1,4 +1,3 @@
-import { dirname, join } from 'path';
 /*
  * Echoes react
  * Copyright (C) 2023-2023 SonarSource SA
```
david-cho-lerat-sonarsource, line 3: should we update to 2024? ;)

gregaubert: oh yeah indeed!

README.md
```diff
@@ -68,6 +68,40 @@ To build the lib, run:
 yarn build
 ```
 
+### Generated files
+
+The lib is built on top of some generated files like the design-tokens of the icons fonts. Those generated files are versionined in the `src/generated` folder.
```
david-cho-lerat-sonarsource, line 73: versionined?

README.md
```diff
@@ -68,6 +68,40 @@ To build the lib, run:
...
+
+#### Generated Icons font
+
+We generated an optimized version of the material-symbold-rounded font that we use for our icons. The whole font contains thousands of icons and weight more than 4Mb, so we generate a subset of it containing only the icons we use in the design system, reducing it's size to a few Kb.
```
david-cho-lerat-sonarsource, line 77: and weight => and weighs
reducing it's size => reducing its size

README.md
gregaubert: You don't know the word?

README.md
```diff
@@ -68,6 +68,40 @@ To build the lib, run:
...
+
+If you add new icons you must run the following command to generate the new optimized font:
+
+```bash
+yarn build-fonts
+```
+
+#### Generated Design Tokens
+
+Our raw design tokens files that are created on Figma are stored inside the `design-tokens/tokens` folder. We can't use these tokens as is, so we transform them using `style-dictionary` to css variable format and store them in the `src/generated` folder. We also generated a few other files to help us with typings.
```
david-cho-lerat-sonarsource, line 87: "we also generated" => "we also generate" ?

README.md
```diff
@@ -68,6 +68,40 @@ To build the lib, run:
...
+
+If new design tokens are added to the raw figma file, we must run the following command to generate the new design tokens css variables:
+
+```bash
+yarn build-tokens
+```
+
+#### Generated i18n keys file
+
+We use the `react-intl` library for internationalization. We store our translations in the `i18n/keys.json` file. This file is generated based on all the translation keys contained in our components. This file if mainly useful for the products that use the lib to know what keys they should translate.
```
david-cho-lerat-sonarsource, line 97: This file **is** mainly..

README.md
```diff
@@ -68,6 +68,40 @@ To build the lib, run:
...
+
+If you add new translations keys, you must run the following command to generate the new `i18n/keys.json` file:
```
david-cho-lerat-sonarsource, line 99: If you add new translation keys,

config/scripts/validate-fonts.sh
```diff
@@ -0,0 +1,11 @@
+#!/bin/bash
+
+yarn build-fonts > /dev/null
+
+if (git diff --quiet src/generated/material-symbols-rounded-optimized.woff2); then
+  echo "Optimized material-symbols font is up to date."
+else
+  echo "Optimized material-symbols font is out of date."
+  echo "Please run 'yarn build-fonts' to update src/generated/material-symbols-rounded-optimized.woff2."
```
david-cho-lerat-sonarsource, line 9: but we just ran `yarn build-fonts`, didn't we? Shouldn't the message say "please commit the updated font" or something?

david-cho-lerat-sonarsource (changes requested):
A  few change requests and a question - otherwise great work :+1:

config/scripts/validate-fonts.sh
gregaubert: In most cases, we will see this in the CI since the validate-fonts task is run as part of the validate-ci task.
So the message will make total sense there, and indeed if you run it yourself locally you'll have one step that's not needed in the message. But I don't think it will happen because if you think of running the check task, you probably run the build command directly instead, and then there is no need for the check task.

gregaubert:
@david-cho-lerat-sonarsource the English errors have been fixed in the last commit, I think everything has been adressed

config/scripts/validate-fonts.sh
david-cho-lerat-sonarsource: Gotcha, thanks.

david-cho-lerat-sonarsource (approved):

PR #90, opened 2024-04-30
Cherry pick commits and prepare version 0.2.2

jeremy-davis-sonarsource:


README.md
```diff
@@ -68,10 +68,49 @@ To build the lib, run:
 yarn build
 ```
 
+### Generated files
+
+The lib is built on top of some generated files like the design-tokens of the icons fonts. Those generated files are versioned in the `src/generated` and `i18n` folders.
```
david-cho-lerat-sonarsource, line 73: `like the design-tokens of the icons fonts`: was this `of` meant to be an `or`?

README.md
```diff
@@ -68,10 +68,49 @@ To build the lib, run:
...
+
+```bash
+yarn build-fonts
+```
+
+#### Generated Design Tokens
+
+Our raw design tokens files that are created on Figma are stored inside the `design-tokens/tokens` folder. We can't use these tokens as is, so we transform them using `style-dictionary` to css variable format and store them in the `src/generated` folder. We also generate a few other files to help us with typings.
+
+If new design tokens are added to the raw figma file, we must run the following command to generate the new design tokens css variables:
```
david-cho-lerat-sonarsource, line 89: NP: `Figma`

david-cho-lerat-sonarsource (approved):
LGTM - just a question about an `of` that maybe should be an `or` in the README :+1:

README.md
jeremy-davis-sonarsource: I'm not sure it's worth fixing here, but definitely on main.
I made a note of it and will create a PR with these

README.md
jeremy-davis-sonarsource: Same as above

PR #92, opened 2024-05-07
NO-JIRA use pre-released sonar-scanner-npm package

lucas-paulger-sonarsource:
use pre-release version of sonar-scanner-npm

### DIFF
--- a/.cirrus.yml
+++ b/.cirrus.yml
@@ -105,7 +105,10 @@ sonar_scan_task:
   depends_on:
     - validation
     - unit_test
+  env:
+    SONAR_SCANNER_NPM_VERSION: 3.6.0-2097
   script:
+    - npm install --no-save "https://$ARTIFACTORY_PRIVATE_USERNAME:$ARTIFACTORY_PRIVATE_PASSWORD@repox.jfrog.io/artifactory/sonarsource-npm-public-dev/sonarqube-scanner/-/sonarqube-scanner-$SONAR_SCANNER_NPM_VERSION.tgz"
     - source .cirrus/cirrus-sq-analysis.sh
 
 promote_task:
--- a/.cirrus.yml
+++ b/.cirrus.yml
@@ -19,8 +19,6 @@ eks_container: &BUILD_TEMPLATE
   namespace: default
   docker_arguments:
     CIRRUS_AWS_ACCOUNT: ${CIRRUS_AWS_ACCOUNT}
-    ARTIFACTORY_PRIVATE_USERNAME: ${ARTIFACTORY_PRIVATE_USERNAME}
-    ARTIFACTORY_PRIVATE_PASSWORD: ENCRYPTED[${ARTIFACTORY_PRIVATE_PASSWORD}]
 
 yarn_cache_template: &YARN_CACHE_TEMPLATE
   yarn_cache:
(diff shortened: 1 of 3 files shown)


.cirrus.yml
```diff
@@ -105,7 +105,10 @@ sonar_scan_task:
   depends_on:
     - validation
     - unit_test
+  env:
+    SONAR_SCANNER_NPM_VERSION: 3.6.0-2082
   script:
+    - npm install --no-save "https://$ARTIFACTORY_PRIVATE_USERNAME:$ARTIFACTORY_PRIVATE_PASSWORD@repox.jfrog.io/artifactory/sonarsource-npm-public-dev/sonarqube-scanner/-/sonarqube-scanner-$SONAR_SCANNER_NPM_VERSION.tgz"
```
lucas-paulger-sonarsource, line 111: this is temporary until the npm package is released.

.cirrus/Dockerfile
lucas-paulger-sonarsource, line 1: I also wanted to try using a node only image
`FROM ${CIRRUS_AWS_ACCOUNT}.dkr.ecr.eu-central-1.amazonaws.com/base:j17-latest`

but it's not clear atm if RE provides this

.cirrus.yml
```diff
@@ -105,7 +105,10 @@ sonar_scan_task:
   depends_on:
     - validation
     - unit_test
+  env:
+    SONAR_SCANNER_NPM_VERSION: 3.6.0-2097
   script:
+    - npm install --no-save "https://$ARTIFACTORY_PRIVATE_USERNAME:$ARTIFACTORY_PRIVATE_PASSWORD@repox.jfrog.io/artifactory/sonarsource-npm-public-dev/sonarqube-scanner/-/sonarqube-scanner-$SONAR_SCANNER_NPM_VERSION.tgz"
```
gregaubert, line 111: Let's do that in the docker file instead? And install it as global maybe?

.cirrus/cirrus-sq-analysis.sh
```diff
@@ -40,6 +40,6 @@ if [[ "${PULL_REQUEST}" ]] || [[ "${GITHUB_BRANCH}" == "main" ]]; then
                                 src/components/icons/**,
                                 src/generated/**,
                                 ")
-  sonar-scanner "${scanner_params[@]}"
+  npx sonar-scanner "${scanner_params[@]}"
```
gregaubert, line 43: does it matter that we run it with npm instead of yarn while all our other commands are run with yarn? I would guess not, just raising the question.

.cirrus.yml
lucas-paulger-sonarsource: Yeah I can do that, I think it aligns more with how our users install it 👍🏻

.cirrus/cirrus-sq-analysis.sh
lucas-paulger-sonarsource: because yarn doesn't have a `--no-save` option, I went with npm. But if I install it globally into docker, I'll use yarn and this won't be an issue 👍🏻

.cirrus.yml
lucas-paulger-sonarsource: docker build task doesn't have access to the artifactory credentials via Vault to have this work. After dogfooding this will work because we will install from npm (not artifactory)

PR #98, opened 2024-05-21
DS-364 Remove aria-label from icons

david-cho-lerat-sonarsource:


jeremy-davis-sonarsource (approved):
Nice!

src/components/links/LinkBase.tsx
```diff
@@ -100,7 +100,7 @@ export const LinkBase = forwardRef<HTMLAnchorElement, LinkProps>((props, ref) =>
         onClick={handleClick}
         ref={ref}>
         {children}
-        {hasExternalIcon && <ExternalIcon />}
+        {hasExternalIcon && <ExternalIcon data-testid="external icon" />}
```
gregaubert, line 103: Should the testids follow a naming convention? I'm mainly thinking about the ones that might leak into the product and could be useful there.
Maybe kebab-case and prefixed with "echoes" and the component that contains it? For this one something like "echoes-link-external-icon".
WDYT ?

gregaubert (approved):

PR #109, opened 2024-06-21
async select

jeremy-davis-sonarsource:


src/components/select/AsyncSelect.tsx
```diff
@@ -0,0 +1,138 @@
...
+
+  const previousQuery = useRef<string>();
+  const valueRef = useRef<string | null>();
+
+  const handleSearch = useCallback(
+    (query: string) => {
+      // Prevent search from being triggered when we select a value
+      const currentValue = getItemLabel(
+        data.find((item) => getItemValue(item) === valueRef.current),
+      );
```
gregaubert, line 77: I don't get that? It's the current label right? We use `getItemLabel`...
But then we compare it with `query` ? `query` is a value or a label? I would think it's the select value that we want to check against right?

src/components/select/AsyncSelect.tsx
```diff
@@ -0,0 +1,138 @@
...
+  id?: string;
+  isDisabled?: boolean;
+  isLoading?: boolean;
+  isNotClearable?: boolean;
+  isRequired?: boolean;
+  labelError?: MantineSelectProps['error'];
+  labelNotFound?: MantineSelectProps['nothingFound'];
+  optionComponent?: MantineSelectProps['itemComponent'];
+  onChange?: MantineSelectProps['onChange'];
+  onSearch?: MantineSelectProps['onSearchChange'];
```
gregaubert, line 45: Should we make it mandatory?

src/components/select/AsyncSelect.tsx
```diff
@@ -0,0 +1,138 @@
...
+  highlight?: SelectHighlight;
+  id?: string;
+  isDisabled?: boolean;
+  isLoading?: boolean;
+  isNotClearable?: boolean;
+  isRequired?: boolean;
+  labelError?: MantineSelectProps['error'];
+  labelNotFound?: MantineSelectProps['nothingFound'];
+  optionComponent?: MantineSelectProps['itemComponent'];
+  onChange?: MantineSelectProps['onChange'];
```
gregaubert, line 44: Should we also make onChange and value mandatory (for normal select too)? since we never really use these components in forms?

src/components/select/AsyncSelect.tsx
```diff
@@ -0,0 +1,138 @@
...
+      if (onChange) {
+        onChange(value);
+      }
+    },
+    [onChange],
+  );
+
+  return (
+    <MantineSelect
+      allowDeselect
```
gregaubert, line 100: we should activate this only we have `!isNotClearable && !isRequired` right? and we should also add it in the normal select I guess?

gregaubert (approved):

src/components/select/AsyncSelect.tsx
jeremy-davis-sonarsource: Good point, but only `!isRequired` I think.

src/components/select/AsyncSelect.tsx
jeremy-davis-sonarsource: `query` in this context is the chosen option's label... :upside_down_face: 
I'm adding a comment

src/components/select/AsyncSelect.tsx
gregaubert: If it should not be clearable then we shouldn't be able to unselect no ?

PR #110, opened 2024-06-21
DS-368 DropdownMenu implementation

david-cho-lerat-sonarsource:


src/components/dropdown-menu/DropdownMenuItemButtonDownload.tsx
```diff
@@ -0,0 +1,37 @@
...
+  DropdownMenuItemButtonBase,
+  DropdownMenuItemButtonBaseProps,
+} from './DropdownMenuItemButtonBase';
+
+interface Props
+  extends Omit<DropdownMenuItemButtonBaseProps, 'isActivable' | 'isActive' | 'prefix'> {}
+
+export function DropdownMenuItemButtonDownload(props: Readonly<Props>) {
+  return (
+    <DropdownMenuItemButtonBase
```
gregaubert, line 32: The download item would actually be a link with some special attributes. I see this in SC:
```
interface ItemDownloadProps extends ListItemProps {
  download: string;
  href: string;
  type?: string;
}

export function ItemDownload(props: Readonly<ItemDownloadProps>) {
  const { children, className, download, href, innerRef, type, ...liProps } = props;
  return (
    <li className="dropdown-menu-download" ref={innerRef} role="none" {...liProps}>
      <ItemDownloadStyled
        className={className}
        download={download}
        href={href}
        rel="noopener noreferrer"
        role="menuitem"
        target="_blank"
        type={type}>
        {children}
      </ItemDownloadStyled>
    </li>
  );
}
```

src/components/dropdown-menu/DropdownMenuItemButtonCopy.tsx
```diff
@@ -0,0 +1,37 @@
...
+import { IconCopy } from '..';
+import {
+  DropdownMenuItemButtonBase,
+  DropdownMenuItemButtonBaseProps,
+} from './DropdownMenuItemButtonBase';
+
+interface Props
+  extends Omit<DropdownMenuItemButtonBaseProps, 'isActivable' | 'isActive' | 'prefix'> {}
+
+export function DropdownMenuItemButtonCopy(props: Readonly<Props>) {
```
gregaubert, line 30: The copy button also has some special behavior, like a copy to clipboard effect some tooltip displayed when copied. On SC I see:

```
interface ItemCopyProps {
  children?: ReactNode;
  className?: string;
  copyValue: string;
}

export function ItemCopy(props: Readonly<ItemCopyProps>) {
  const { children, className, copyValue } = props;
  const [copySuccess, handleCopy] = useCopyClipboardEffect(copyValue);
  return (
    <Tooltip content={translate('copied_action')} isOpen={copySuccess} side={TooltipSide.Left}>
      <li className="dropdown-menu-copy" role="none">
        <ItemButtonStyled
          className={classNames('fs-exclude', className)}
          onClick={handleCopy}
          role="menuitem">
          {children}
        </ItemButtonStyled>
      </li>
    </Tooltip>
  );
}
```

src/components/dropdown-menu/DropdownMenuItemBase.tsx
```diff
@@ -0,0 +1,181 @@
...
+
+export type DropdownMenuItemBaseProps = CheckProps & {
+  ariaLabel?: string;
+  children: ReactNode;
+  helpText?: JSX.Element | string;
+  isDisabled?: boolean;
+  isDestructive?: boolean;
+  onClick?: MouseEventHandler<HTMLDivElement>;
+  prefix?: JSX.Element;
+  suffix?: JSX.Element;
```
gregaubert, line 44: I used `ReactNode` for the prefix and suffix in select, I wonder what would be the reason to restrict to `JSX.Element`?

src/components/dropdown-menu/DropdownMenuItemBase.tsx
```diff
@@ -0,0 +1,181 @@
...
+};
+
+export function DropdownMenuItemBase({
+  ariaLabel,
+  children,
+  helpText,
+  isCheckable,
+  isChecked,
+  isDisabled,
+  isDestructive,
```
gregaubert, line 54: Let's set default values here, to avoid relying on a third `undefined` state in the implem.

src/components/dropdown-menu/DropdownMenuGroupLabel.tsx
```diff
@@ -0,0 +1,38 @@
...
+
+import styled from '@emotion/styled';
+import * as radixDropdownMenu from '@radix-ui/react-dropdown-menu';
+import { PropsWithChildren } from 'react';
+
+export function DropdownMenuGroupLabel({
+  ariaLabel,
+  children,
+}: Readonly<PropsWithChildren<{ ariaLabel?: string }>>) {
+  return <StyledGroupLabel aria-label={ariaLabel}>{children}</StyledGroupLabel>;
```
gregaubert, line 29: I don't think a group label needs an aria-label, it's not an interactive element, it's content should be enough

src/components/dropdown-menu/DropdownMenuGroupLabel.tsx
```diff
@@ -0,0 +1,38 @@
...
+}
+
+const StyledGroupLabel = styled(radixDropdownMenu.Label)`
+  color: var(--echoes-color-text-default);
+  font-size: var(--echoes-font-size-body-small);
+  font-weight: var(--echoes-font-weight-semi-bold);
```
gregaubert, line 35: Let's use the correct font directly instead of reacreating it with different props. Something like : `font: var(--echoes-typography-paragraph-small-semi-bold);`

src/components/dropdown-menu/DropdownMenuItemBase.tsx
```diff
@@ -0,0 +1,183 @@
...
+    );
+  }
+
+  return (
+    <StyledDropdownMenuItemBase
+      disabled={isDisabled}
+      isCheckable={isCheckable}
+      isChecked={isChecked}
+      isDestructive={isDestructive}
+      onClick={isDisabled ? undefined : onClick}>
```
gregaubert, line 75: Should we just rely on radix to handle it correctly ?

src/components/dropdown-menu/DropdownMenuItemBase.tsx
```diff
@@ -0,0 +1,183 @@
...
+      <StyledLeftHandSide>
+        {checkMarkOrPlaceHolder}
+
+        {prefix}
+
+        <StyledLabelAndHelpText aria-label={ariaLabel}>
```
gregaubert, line 81: The aria label should be on the `StyledDropdownMenuItemBase` if we have one, that's the interactive part. But I wonder if it's really needed?

src/components/dropdown-menu/DropdownMenuItemBase.tsx
```diff
@@ -0,0 +1,183 @@
...
+  display: flex;
+  flex-direction: column;
+  justify-content: space-between;
+  row-gap: var(---echoes-dimension-space-25);
+`;
+
+const StyledHelpText = styled.span<{ isDisabled?: boolean }>`
+  color: ${(props) =>
+    props.isDisabled ? 'var(--echoes-color-text-disabled)' : 'var(--echoes-color-text-subdued)'};
+  font-size: var(--echoes-font-size-body-small);
```
gregaubert, line 112: Let's set the font directly with the font design token

src/components/dropdown-menu/DropdownMenuItemBase.tsx
```diff
@@ -0,0 +1,183 @@
...
+`;
+
+const StyledSuffix = styled.span`
+  align-items: center;
+  color: var(--echoes-color-text-subdued);
+  display: flex;
+`;
+
+const StyledDropdownMenuItemBase = styled(radixDropdownMenu.Item)<
+  Pick<DropdownMenuItemBaseProps, 'isCheckable' | 'isChecked' | 'isDestructive'>
```
gregaubert, line 122: Seems like `isCheckable` isn't used

src/components/dropdown-menu/DropdownMenuItemBase.tsx
```diff
@@ -0,0 +1,183 @@
...
+>`
+  align-items: center;
+  background-color: ${(props) =>
+    props.isChecked
+      ? 'var(--echoes-color-background-selected-weak-default)'
+      : 'var(--echoes-color-background-default)'};
+  color: ${(props) =>
+    props.isDestructive ? 'var(--echoes-color-text-danger)' : 'var(--echoes-color-text-default)'};
+  column-gap: var(--echoes-dimension-space-100);
+  cursor: pointer;
```
gregaubert, line 132: I don't think we should force a `cursor: pointer`

src/components/dropdown-menu/DropdownMenuItemBase.tsx
```diff
@@ -0,0 +1,183 @@
...
+      return props.isChecked
+        ? 'var(--echoes-color-background-selected-weak-pressed)'
+        : 'var(--echoes-color-background-default-active)';
+    }};
+  }
+
+  /* when the current URL matches 'to', react-router adds an 'active' class to the parent 'a' */
+  .active & {
+    background-color: var(--echoes-color-background-default-active);
+  }
```
gregaubert, line 173: Should we have this styling only in the link type item ?

src/components/dropdown-menu/DropdownMenuItemBase.tsx
```diff
@@ -0,0 +1,183 @@
...
+    border-radius: 0;
+    outline: none;
+  }
+
+  /* when the item is clicked */
+  &:active {
+    background-color: ${(props) => {
+      if (props.isDestructive) {
+        return 'var(--echoes-color-background-danger-weak-active)';
+      }
```
gregaubert, line 162: Should the specific destructive styling be moved to the destructive type item ?

src/components/dropdown-menu/DropdownMenuItemButtonDefault.tsx
```diff
@@ -0,0 +1,27 @@
...
+ * You should have received a copy of the GNU Lesser General Public License
+ * along with this program; if not, write to the Free Software Foundation,
+ * Inc., 51 Franklin Street, Fifth Floor, Boston, MA  02110-1301, USA.
+ */
+
+import { DropdownMenuItemBase, DropdownMenuItemBaseProps } from './DropdownMenuItemBase';
+
+type Props = Omit<DropdownMenuItemBaseProps, 'isCheckable' | 'isChecked' | 'isDestructive'>;
+
+export function DropdownMenuItemButtonDefault(props: Readonly<Props>) {
```
gregaubert, line 25: Let's name it `DropdownMenuItemButton` ?

src/components/tooltip/Tooltip.tsx
```diff
@@ -68,7 +68,7 @@ const ARROW_PADDING = 12;
  *
  * Since the tooltips are appended to the body, they are in the root Stacking Context. If other elements are also there, the z-index will determine which appears on top. By creating a new Stacking Context for your app, it ensures that z-indexed elements will stay within that context, while tooltips will be painted on top, in the parent Stacking Context.
  */
-export const Tooltip = forwardRef<HTMLDivElement, Props>((props, ref) => {
+export const Tooltip = forwardRef<HTMLButtonElement, Props>((props, ref) => {
```
gregaubert, line 71: Should we actually be more generic with `HTMLElement` ? since we don't know if it's really gonna be a button ?

src/components/dropdown-menu/index.ts
```diff
@@ -0,0 +1,29 @@
...
+
+export { DropdownMenu } from './DropdownMenu';
+export { DropdownMenuGroupLabel } from './DropdownMenuGroupLabel';
+export { DropdownMenuItemButtonCheckable } from './DropdownMenuItemButtonCheckable';
+export { DropdownMenuItemButtonCopy } from './DropdownMenuItemButtonCopy';
+export { DropdownMenuItemButtonDefault } from './DropdownMenuItemButtonDefault';
+export { DropdownMenuItemButtonDestructive } from './DropdownMenuItemButtonDestructive';
+export { DropdownMenuItemButtonDownload } from './DropdownMenuItemButtonDownload';
+export { DropdownMenuItemLink } from './DropdownMenuItemLink';
+export { DropdownMenuSeparator } from './DropdownMenuSeparator';
```
gregaubert, line 29: @jeremy-davis-sonarsource @david-cho-lerat-sonarsource What do you guys think of the way of doing of radix? To basically attach all the sub-component to the parent component so that you can just import the parent component and then do stuff like this:
```ts
<DropdownMenu items={
  <>
    <DropdownMenu.GroupLabel/>
    <DropdownMenu.ItemButton/>
  </>
}>
```

src/components/dropdown-menu/DropdownMenuSeparator.tsx
```diff
@@ -0,0 +1,33 @@
...
+
+import styled from '@emotion/styled';
+import * as radixDropdownMenu from '@radix-ui/react-dropdown-menu';
+
+export function DropdownMenuSeparator() {
+  return <StyledSeparator />;
+}
+
+const StyledSeparator = styled(radixDropdownMenu.Separator)`
+  border-top: 1px solid var(--echoes-color-border-weak);
```
gregaubert, line 29: let's use the design-token for the border width

src/components/dropdown-menu/DropdownMenuItemLink.tsx
```diff
@@ -0,0 +1,60 @@
...
+    </StyledLinkBase>
+  );
+}
+
+const StyledLinkBase = styled(LinkBase)`
+  text-decoration: none;
+`;
+
+const StyledIconLinkExternal = styled(IconLinkExternal)`
+  font-size: var(--echoes-font-size-body-medium);
```
gregaubert, line 59: Let's define the full font

src/components/dropdown-menu/DropdownMenuItemLink.tsx
```diff
@@ -0,0 +1,60 @@
...
+export function DropdownMenuItemLink({
+  isDisabled,
+  isExternal: isExternalProp = false,
+  to,
+  ...props
+}: Readonly<Props>) {
+  const toAsString =
+    typeof to === 'string' ? to : `${to.pathname ?? ''}${to.search ?? ''}${to.hash ?? ''}`;
+
+  const isExternal = isExternalProp || toAsString.startsWith('http');
```
gregaubert, line 41: This logic is already done in the LinkBase right no need to repeat it?

src/components/links/LinkBase.tsx
```diff
@@ -113,9 +122,9 @@ export const LinkBase = forwardRef<HTMLAnchorElement, LinkProps>((props, ref) =>
   }
 
   return (
-    <RouterLink {...restAndRadixProps} onClick={handleClick} ref={ref} to={to}>
+    <RouterNavLink end {...restAndRadixProps} onClick={handleClick} ref={ref} style={style} to={to}>
```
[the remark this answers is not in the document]
gregaubert: Maybe we can add an optional prop to LinkBase to be able to swap the type of link (RouterLink vs RouterNavLink)?
So that we can keep the same behavior and only use NavLink where it's needed?

src/components/dropdown-menu/DropdownMenuItemLink.tsx
david-cho-lerat-sonarsource: yes, there is, it's so we can place `IconLinkExternal` in the `suffix` slot if needed

gregaubert: Oh! ok got it, makes sense!

src/components/dropdown-menu/DropdownMenuItemBase.tsx
david-cho-lerat-sonarsource: no reason, changing to `ReactNode`

src/components/dropdown-menu/DropdownMenuItemBase.tsx
david-cho-lerat-sonarsource: that's what I tried at first, but I was able to click disabled elements

src/components/dropdown-menu/DropdownMenuItemBase.tsx
david-cho-lerat-sonarsource: why not? All items in a dropdown are supposed to be clickable if they're not disabled (or labels, separators,..)
If we don't add it here then every "customer" will have to add it themselves

src/components/dropdown-menu/DropdownMenuItemButtonDefault.tsx
david-cho-lerat-sonarsource: yes I was thinking the same thing yesterday

src/components/tooltip/Tooltip.tsx
david-cho-lerat-sonarsource: `radixTooltip.Trigger` specifically expects an `HTMLButtonElement` ref

david-cho-lerat-sonarsource: I will use `HTMLElement` here and cast to `HTMLButtonElement` when used within `radixTooltip.Trigger`

src/components/dropdown-menu/DropdownMenuItemBase.tsx
gregaubert: Ok let's keep it like this then!

src/components/tooltip/Tooltip.tsx
gregaubert: It's weird that radix trigger only accept HTMLButtonElement ref while it's opened to be used with any kind of children

src/components/dropdown-menu/DropdownMenuItemBase.tsx
david-cho-lerat-sonarsource: but this is what we do everywhere - what other cursor should be used? Some items will be links, some buttons - should we have different cursors?

design-tokens/tokens/layer1/base.json
```diff
@@ -112,6 +112,10 @@
           "value": "4rem",
           "type": "dimension"
         },
+        "1875": {
+          "value": "9.375rem",
+          "type": "dimension"
+        },
```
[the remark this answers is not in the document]
david-cho-lerat-sonarsource: Discussed with @marciopmoreira6 , we've added the level 3 tokens as suggested

src/components/dropdown-menu/DropdownMenuItemBase.tsx
david-cho-lerat-sonarsource: let's keep it for now, as discussed. @marciopmoreira6 thinks we should keep it to, JIRA does the same thing :shrug:

src/components/dropdown-menu/DropdownMenuItemButtonDownload.tsx
david-cho-lerat-sonarsource: cutting this branch for now, the special behavior will be implemented later

src/components/dropdown-menu/DropdownMenuItemButtonCopy.tsx
david-cho-lerat-sonarsource: cutting this branch for now, the special behavior will be implemented later

stories/DropdownMenuItemButtonCheckable-stories.tsx
```diff
@@ -0,0 +1,292 @@
...
+ */
+
+import styled from '@emotion/styled';
+import type { Meta, StoryObj } from '@storybook/react';
+import { PropsWithChildren, forwardRef } from 'react';
+import { DropdownMenu, IconChevronDown } from '../src';
+
+const meta: Meta<typeof DropdownMenu.ItemButtonCheckable> = {
+  component: DropdownMenu.ItemButton,
+  title: 'DropdownMenu.ItemButtonCheckable',
```
jeremy-davis-sonarsource, line 28: You could replace the dot with a slash in the title. This will create a kind of category in storybook:
image

stories/DropdownMenuItemButtonCheckable-stories.tsx
```diff
@@ -0,0 +1,292 @@
...
+  render: ({ children, ...args }) => (
+    <BasicWrapper>
+      <DropdownMenu.Root
+        isOpen
+        items={
+          <DropdownMenu.ItemButtonCheckable {...args}>{children}</DropdownMenu.ItemButtonCheckable>
+        }>
+        <MenuButton />
+      </DropdownMenu.Root>
+    </BasicWrapper>
```
jeremy-davis-sonarsource, line 74: I'd move a maximum to the wrapper, to make it DRYer

jeremy-davis-sonarsource (approved):
I don't think I'd have created this many stories, so kudos!

PR #112, opened 2024-06-25
[NO JIRA] Add dimension-1875 token (150px) for the dropdown panel min-width

david-cho-lerat-sonarsource:


design-tokens/tokens/layer1/base.json
```diff
@@ -112,6 +112,10 @@
           "value": "4rem",
           "type": "dimension"
         },
+        "1875": {
+          "value": "9.375rem",
+          "type": "dimension"
+        },
```
gregaubert, line 118: This value seems so random...
This is 150px right ?
@marciopmoreira6 why do we define width size for input in px and in rem for dropdown?
Also these two new values are specific for dropdowns right? shouldn't we add them in layer 3 instead?

david-cho-lerat-sonarsource: yes it does look random. I suggested `dimension-2000` but @marciopmoreira6 said that if we wanted it to be 150px we might as well give the precise value

gregaubert: let's sync with @marciopmoreira6 we just created input size tokens yesterday as pixels values, I think we should have the something aligned between input sizes and dropdowns.

PR #114, opened 2024-06-26
Change the structure and states of some sentiment colors like accent and danger

marciopmoreira6:


gregaubert:
@marciopmoreira6 @david-cho-lerat-sonarsource we need to start creating tickets for this kind of tasks where we rename design tokens. It can be a "breaking change" so we want that kind of changes to appear in the release note. If we don't have a ticket, it won't show up in the release note.

gregaubert (approved):
Good to merge, I could not find any usage anywhere.

marciopmoreira6:
Thanks, @gregaubert. I created a task with those changes and added it to 0.4.0

PR #115, opened 2024-06-27
DS-369 Add styling to Select and SelectAsync

gregaubert:


src/components/select/SelectAsync.tsx
```diff
@@ -109,41 +100,56 @@ export const SelectAsync = forwardRef<HTMLInputElement, PropsWithLabels<Props>>(
...
+        isLoading,
+        isClearable,
+      })}
       searchable
-      value={value}
+      value={value ?? null} // Mantine only clears the value if `null`, not `undefined`
       variant={highlight}
+      {...selectProps}
     />
   );
```
jeremy-davis-sonarsource, line 140: This is mostly shared between the two, I wonder if it isn't better to have a `SelectBase` ?

jeremy-davis-sonarsource (approved):
LGTM, just a suggestion to make it more DRY

jeremy-davis-sonarsource (approved):
Awesome!

PR #116, opened 2024-06-28
DS-369 Add DropdownMenu, Select and SelectAsync components

gregaubert:


src/components/select/SelectItemCommons.tsx
```diff
@@ -0,0 +1,215 @@
...
+
+    &[data-selected]::after {
+      background-color: var(--echoes-color-icon-disabled);
+      border-color: var(--echoes-color-background-disabled);
+    }
+  }
+`;
+SelectItemSatusRadio.displayName = 'SelectItemSatusRadio';
+
+/**
```
david-cho-lerat-sonarsource, line 197: do we want to keep this comment block as is?

src/components/select/SelectCommons.tsx
```diff
@@ -0,0 +1,329 @@
...
+  hasError?: boolean;
+  highlight?: SelectHighlight;
+  isDisabled?: boolean;
+  isLoading?: boolean;
+  isNotClearable?: boolean;
+  isSearchable?: boolean;
+  isRequired?: boolean;
+  labelError?: MantineSelectProps['error'];
+  labelNotFound?: MantineSelectProps['nothingFound'];
+  limit?: MantineSelectProps['limit']; // might change for a max height
```
david-cho-lerat-sonarsource, line 47: so which is it? limit or max height? We probably shouldn't leave this comment here

src/components/select/SelectCommons.tsx
```diff
@@ -0,0 +1,329 @@
...
+      valueIcon,
+      ...selectProps
+    } = props;
+
+    const intl = useIntl();
+
+    const itemComponent = useSelectItemComponent(optionComponent, optionType);
+    const isClearable = !isNotClearable && !isRequired;
+
+    // TODO Highlighter for search
```
david-cho-lerat-sonarsource, line 91: if it's not done now it should probably be a ticket instead of a TODO comment

src/components/select/SelectCommons.tsx
```diff
@@ -0,0 +1,329 @@
...
+
+  // Error element styles, defined by error prop
+  & .mantine-Select-error {
+    font: var(--echoes-typography-paragraph-default-regular);
+    color: var(--echoes-color-text-danger);
+
+    margin-top: var(--echoes-dimension-space-100);
+  }
+
+  // Main input element, with styling for the default and ghost highlight and the difference states
```
david-cho-lerat-sonarsource, line 182: "and the different states"?

src/components/select/SelectCommons.tsx
jeremy-davis-sonarsource: https://sonarsource.atlassian.net/browse/DS-441 :+1:

jeremy-davis-sonarsource (approved):

PR #119, opened 2024-07-04
DS-368 DropdownMenu fixes

david-cho-lerat-sonarsource:


src/components/links/LinkBase.tsx
```diff
@@ -30,14 +30,7 @@ import {
...
 
-type RouterNavLinkPropsAllowed =
-  | 'download'
-  | 'end'
-  | 'reloadDocument'
-  | 'state'
-  | 'style'
-  | 'title'
-  | 'to';
+type RouterNavLinkPropsAllowed = 'download' | 'reloadDocument' | 'state' | 'style' | 'title' | 'to';
```
david-cho-lerat-sonarsource, line 33: @jeremy-davis-sonarsource this would be a "breaking change" in that users would now have to pass `matchesExactPath` instead of `end`, and the logic is reversed (defaults to `true) - WDYT?

jeremy-davis-sonarsource: It should default to false and be called something like `isMatchingExactPath`:
https://docs.google.com/document/d/1r50MkSdJEznh8ijjpgRoJffhQ9IKj5tM41Y2XbwEmO0/edit#heading=h.n71kn868u7ey

david-cho-lerat-sonarsource: OK, I've renamed and defaulted to `false` in `LinkBase`. I've defaulted to `true` in `DropdownMenuItemLink`, however, because it looks like it's the more common case. WDYT?

jeremy-davis-sonarsource: Then I'd call it `isMatchingPartialPath` and make it default to false.

david-cho-lerat-sonarsource: reverting this to default to a partial match like react-router, as discussed

src/components/dropdown-menu/DropdownMenuItemBase.tsx
```diff
@@ -60,39 +60,50 @@ export function DropdownMenuItemBase({
...
 }
 
+const StyledIconCheck = styled(IconCheck)`
+  font-size: var(--echoes-font-size-body-medium);
+`;
+
 const CheckmarkPlaceholder = styled(IconCheck)`
+  font-size: var(--echoes-font-size-body-medium);
   visibility: hidden;
 `;
```
jeremy-davis-sonarsource, line 99: NP: you could reuse the StyledIconCheck instead, so that any changes to it doesn't need to be reproduced here.

jeremy-davis-sonarsource (approved):
:muscle:

PR #130, opened 2024-07-12
DS-494 Use enum instead of string for DropdownMenu panel position

david-cho-lerat-sonarsource:


src/components/dropdown-menu/DropdownMenu.tsx
```diff
@@ -128,6 +134,7 @@ const DropdownMenuRoot = forwardRef<HTMLButtonElement, DropdownMenuRootProps>(
 DropdownMenuRoot.displayName = 'DropdownMenu.Root';
 
 export const DropdownMenu = {
+  Align,
   GroupLabel: DropdownMenuGroupLabel,
   ItemButton: DropdownMenuItemButton,
   ItemButtonCheckable: DropdownMenuItemButtonCheckable,
```
jeremy-davis-sonarsource, line 140: I think it would be better to export it separately, as `DropdownMenuAlign`:
The typing indicates `Align | undefined`, so it's hard to guess that you should like inside `DropdownMenu`.
And this way we separate clearly what are basically sub-components from types, enums, etc..

WDYT?
(curious about Greg's point of view too)

david-cho-lerat-sonarsource: makes sense to me, changed it (there's still time to change back depending on Greg's opinion ;))

gregaubert: Yeah agree, let's export it separately

jeremy-davis-sonarsource (approved):
Nice!

PR #131, opened 2024-07-12
DS-499 Allow use of `suffix` slot in DropdownMenu.ItemLink (for badges etc)

david-cho-lerat-sonarsource:


src/components/dropdown-menu/DropdownMenuItemLink.tsx
```diff
@@ -38,18 +39,27 @@ export function DropdownMenuItemLink({
...
+  let composedSuffix = suffix ?? externalSuffix;
+
+  if (suffix && externalSuffix) {
+    composedSuffix = (
+      <span className="sw-flex sw-gap-2 sw-items-center">
+        {suffix}
+        {externalSuffix}
+      </span>
+    );
+  }
```
jeremy-davis-sonarsource, line 53: NP: I'd move this to a component or helper function

src/components/dropdown-menu/DropdownMenuItemLink.tsx
```diff
@@ -38,18 +39,27 @@ export function DropdownMenuItemLink({
 
   const isExternal = isExternalProp || toAsString.startsWith('http');
 
+  const externalSuffix = isExternal ? <StyledIconLinkExternal /> : undefined;
+
+  let composedSuffix = suffix ?? externalSuffix;
+
+  if (suffix && externalSuffix) {
+    composedSuffix = (
+      <span className="sw-flex sw-gap-2 sw-items-center">
```
jeremy-davis-sonarsource, line 48: We don't have tailwind in echoes

jeremy-davis-sonarsource (changes requested):
You should a test and a story for this new use case

src/components/dropdown-menu/DropdownMenuItemLink.tsx
david-cho-lerat-sonarsource: hm how come it's working, then? :thinking: Because it's using the consumer's tailwind?

jeremy-davis-sonarsource: yeah, it works because those classes are used in the product so they are included, but it doesn't work in storybook for instance (and it means the consumer must include tailwind with the `sw-` prefix configured)

david-cho-lerat-sonarsource: changed to styled component

src/components/dropdown-menu/DropdownMenuItemLink.tsx
david-cho-lerat-sonarsource: done

src/components/dropdown-menu/DropdownMenuItemLink.tsx
```diff
@@ -23,13 +23,29 @@ import { IconLinkExternal } from '../icons/IconLinkExternal';
 import { LinkBase, LinkProps } from '../links/LinkBase';
 import { DropdownMenuItemBase, DropdownMenuItemBaseProps } from './DropdownMenuItemBase';
 
-type Props = Omit<DropdownMenuItemBaseProps, 'isCheckable' | 'isChecked' | 'suffix'> &
+type Props = Omit<DropdownMenuItemBaseProps, 'isCheckable' | 'isChecked'> &
   Pick<LinkProps, 'isExternal' | 'isMatchingFullPath' | 'to'>;
 
+const composedSuffix = ({ isExternal, suffix }: Readonly<Partial<Props>>) => {
```
jeremy-davis-sonarsource, line 29: NP: I'd  `Pick`  specifically what is needed

jeremy-davis-sonarsource (commented):
still no tests? :(

jeremy-davis-sonarsource (approved):
:muscle:

PR #134, opened 2024-07-15
DS-508 Implement Popover

jeremy-davis-sonarsource:


src/components/popover/Popover.tsx
```diff
@@ -0,0 +1,145 @@
...
+ *
+ * ```CSS
+ *   position: relative;
+ *   z-index: 0;
+ * ```
+ *
+ * Since the popovers are appended to the body, they are in the root Stacking Context. If other elements are also there, the z-index will determine which appears on top. By creating a new Stacking Context for your app, it ensures that z-indexed elements will stay within that context, while popovers will be painted on top, in the parent Stacking Context.
+ */
+export const Popover = forwardRef<HTMLButtonElement, Props>((props, ref) => {
+  const { align, children, description, extraContent, footer, title, isOpen, side, ...radixProps } =
```
david-cho-lerat-sonarsource, line 74: OCD NP: isOpen and side are not sorted :grin:

david-cho-lerat-sonarsource (approved):
Nice! :muscle:

src/components/popover/Popover.tsx
```diff
@@ -0,0 +1,145 @@
...
+`;
+
+const PopoverFooter = styled.div`
+  margin-top: var(--echoes-dimension-space-200);
+`;
+
+const PopoverContent = styled(RadixPopover.Content)`
+  border: var(--echoes-border-width-default) solid var(--echoes-color-border-weak);
+  border-radius: var(--echoes-border-radius-400);
+  padding: var(--echoes-dimension-space-300) var(--echoes-dimension-space-250);
```
david-cho-lerat-sonarsource, line 127: (this looks huge to me ..)

src/components/popover/Popover.tsx
```diff
@@ -0,0 +1,145 @@
...
+/* This is the distance between the point of the arrow and the trigger */
+const POPOVER_OFFSET = 4;
+
+/* This is the padding between the edge of the tooltip and the arrow. */
+const ARROW_PADDING = 16;
+
+/**
+ * **Popovers must be attached to a button to be accessible.**
+ *
+ * ### Stacking Context
```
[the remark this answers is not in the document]
jeremy-davis-sonarsource: Good question! We could add a dedicated page to the migration guide section and point to it from every overlay-type component.

PR #137, opened 2024-07-16
DS-512 Fix a11y issue with DropdownMenuItemLink

david-cho-lerat-sonarsource:


src/components/dropdown-menu/DropdownMenuItemBase.tsx
```diff
@@ -85,10 +84,34 @@ export function DropdownMenuItemBase({
...
+        {itemContents}
+      </StyledLinkBase>
+    ) : (
+      itemContents
+    );
+
+  return (
+    <StyledRadixDropdownMenuItem
+      aria-label={ariaLabel}
+      {...(linkProps !== undefined ? { asChild: true } : {})}
```
gregaubert, line 102: Works fine.

I'm nitpicking a bit, but the only thing I don't like is that now we have link item logic leaking into the base and every other item components.

Based on the same principle, maybe we can instead expose some internal props named `asChildWrapper` and `asChildWrapperProps` prop instead? It would behave the same way as you did, setting up the `asChild` as true. It would keep things generic in the base, and only the items that need that can be used, no need to change the other items.

To have the internal props, you might need to split `DropdownMenuItemBaseProps` in two: one that's exported with the public props and one that's not exported and only to be used by our lib. It would contain the' asChildWrapper, ' for example, or any other prop that should be public.

src/components/dropdown-menu/DropdownMenuItemBase.tsx
```diff
@@ -66,25 +72,38 @@ export function DropdownMenuItemBase({
     );
   }
 
-  return (
-    <StyledRadixDropdownMenuItem
-      aria-label={ariaLabel}
-      className={className}
-      disabled={isDisabled}
-      onClick={isDisabled ? undefined : onClick}>
+  const getStyledItemContents = ({ label }: { label: ReactNode }) => (
```
gregaubert, line 75: do we need a usecallback?

david-cho-lerat-sonarsource: added

gregaubert (approved):

PR #139, opened 2024-07-16
DS-513 Added heights in Echoes + negative spacing

marciopmoreira6:


design-tokens/tokens/layer1/base.json
```diff
@@ -55,7 +55,7 @@
           "type": "dimension"
         }
       },
-      "size": {
+      "width": {
```
gregaubert, line 58: ouf this is a big naming change, it's used everywhere... do we really need to make a difference between width and height on layer 1 ?

design-tokens/tokens/layer3/base.json
```diff
@@ -4,7 +4,7 @@
       "overlays": {
         "maxWidth": {
           "default": {
-            "value": "{echoes.dimension.size.5000}",
+            "value": "{echoes.dimension.width.5000}",
```
gregaubert, line 7: I thought you wanted to add a layer 3 design token for the button heights

design-tokens/tokens/layer1/base.json
```diff
@@ -116,6 +116,48 @@
...
+        },
+        "1000": {
+          "value": "2.5rem",
+          "type": "dimension"
+        }
+      },
+      "negative": {
+        "spacing": {
+          "100": {
+            "value": "-0.063rem",
```
gregaubert, line 157: you want 1 px ?

marciopmoreira6: we can change to -1px instead of the -0.063rem

design-tokens/tokens/layer3/base.json
marciopmoreira6: Fixed

design-tokens/tokens/layer1/base.json
marciopmoreira6: Discussed live with @gregaubert

design-tokens/tokens/layer1/base.json
jeremy-davis-sonarsource: Yeah, I don't think negative spacing like this is related to font size

design-tokens/tokens/layer1/base.json
```diff
@@ -116,6 +116,48 @@
...
+      },
+      "height": {
+        "100": {
+          "value": "0.25rem",
+          "type": "dimension"
+        },
+        "200": {
+          "value": "0.5rem",
+          "type": "dimension"
+        },
```
jeremy-davis-sonarsource, line 128: We're missing some values:
RadioButtonGroup's "SelectionIndicator" is using `echoes-dimension-size-75` for its height (so we're missing 150)

jeremy-davis-sonarsource (changes requested):
What about border sizes?
`width/height` make no sense. Should we use `space` instead?
==> I've got a `size-50` border to switch out

design-tokens/tokens/layer1/base.json
marciopmoreira6: We have a specific layer for "Border Widths" outside of the "Sizes" layer

marciopmoreira6: Ok, I've just added the 150 value

jeremy-davis-sonarsource: Then we're missing the right border width for the Radiobutton in the Selects, we only have the default 1px currently

src/components/buttons/ButtonStyles.tsx
```diff
@@ -165,10 +165,10 @@ export const BUTTON_VARIETY_STYLES = {
 export const BUTTON_SIZE_STYLE = {
   [ButtonSize.Medium]: {
     '--button-padding': 'var(--echoes-dimension-space-100)',
-    '--button-height': 'var(--echoes-dimension-size-400)',
+    '--button-height': 'var(--echoes-dimension-height-800)',
```
gregaubert, line 168: Let's use the new layer 3 buttons-medium and buttons-large

gregaubert (approved):
Looks good let's just use the layer3 on buttons

jeremy-davis-sonarsource (approved):

PR #141, opened 2024-07-17
DS-500 Add Button component

gregaubert:


src/components/spinner/Spinner.tsx
```diff
@@ -77,33 +80,42 @@ export const Spinner = forwardRef<HTMLSpanElement, Props>((props, ref) => {
...
 
-const SpinnerWrapper = styled.span<{ inline: boolean }>`
+const SpinnerWrapper = styled.span<{ inline: boolean; isLoading?: boolean }>`
   display: ${(props) => (props.inline ? 'inline-block' : 'block')};
   position: relative;
-  ${(props) => (props.inline ? 'margin: 0 var(--echoes-dimension-space-50);' : '')}
+  ${(props) =>
+    props.inline && props.isLoading ? 'margin: 0 var(--echoes-dimension-space-50);' : ''}
+
+  &.echoes-button-spinner-wrapper {
```
gregaubert, line 96: I'm not super happy with this pattern, I wanted to use `${ButtonStyled} & {}` instead but ran into some issues that I didn't understand so I tried to move faster by going around the issue.

src/components/buttons/ButtonTypes.ts
```diff
@@ -0,0 +1,41 @@
...
+ * You should have received a copy of the GNU Lesser General Public License
+ * along with this program; if not, write to the Free Software Foundation,
+ * Inc., 51 Franklin Street, Fifth Floor, Boston, MA  02110-1301, USA.
+ */
+
+import { ButtonHTMLAttributes } from 'react';
+
+export type HTMLButtonProps = Pick<
+  ButtonHTMLAttributes<HTMLButtonElement>,
+  'id' | 'name' | 'role' | 'style' | 'type' | 'form'
```
david-cho-lerat-sonarsource, line 25: OCD NP: 'form' first? :grin:

src/components/buttons/__tests__/Button-test.tsx
```diff
@@ -0,0 +1,134 @@
...
+
+it('should not call onClick function when disabled', async () => {
+  const onClick = jest.fn();
+
+  const { user } = render(
+    <Button isDisabled onClick={onClick}>
+      Click me
+    </Button>,
+    {},
+    // We skip the pointer-events:none check from user-event to be able to test clicking on the disabled checkbox
```
david-cho-lerat-sonarsource, line 41:
```suggestion
    // We skip the pointer-events:none check from user-event to be able to test clicking on the disabled button
```

src/components/buttons/__tests__/Button-test.tsx
```diff
@@ -0,0 +1,134 @@
...
+    { pointerEventsCheck: PointerEventsCheckLevel.Never },
+  );
+
+  expect(screen.getByRole('button', { name: 'Click me' })).toBeDisabled();
+
+  await user.click(screen.getByRole('button', { name: 'Click me' }));
+  expect(onClick).not.toHaveBeenCalled();
+});
+
+it("should show a loading state, it doens't prevent clicking", () => {
```
david-cho-lerat-sonarsource, line 51:
```suggestion
it("should show a loading state, it doesn't prevent clicking", () => {
```

src/components/buttons/Button.tsx
```diff
@@ -0,0 +1,111 @@
...
+          ...BUTTON_VARIETY_STYLES[variety],
+          ...BUTTON_SIZE_STYLE[size],
+        }),
+        [variety, size],
+      )}
+      disabled={isDisabled}
+      onClick={handleClick}
+      ref={ref}
+      type={type}>
+      {isDefined(isLoading) && (
```
david-cho-lerat-sonarsource, line 95: it'd be nice to have a story showcasing the loading state directly instead of having to think about toggling isLoading in SB - WDYT?

stories/Button-stories.tsx
```diff
@@ -0,0 +1,82 @@
...
+    size: { control: { type: 'select' }, options: Object.values(ButtonSize) },
+    variety: { control: { type: 'select' }, options: Object.values(ButtonVariety) },
+  },
+};
+
+export default meta;
+
+type Story = StoryObj<typeof Button>;
+
+export const Default: Story = {
```
david-cho-lerat-sonarsource, line 39: maybe you can use the [BasicWrapper](https://github.com/SonarSource/echoes-react/blob/main/stories/DropdownMenu-stories.tsx#L117) we use in other stories so that the component isn't on the top-left?

david-cho-lerat-sonarsource (changes requested):

src/components/buttons/ButtonStyles.tsx
```diff
@@ -0,0 +1,164 @@
...
+    border: none;
+
+    cursor: not-allowed;
+    pointer-events: none;
+  }
+`;
+
+ButtonStyled.displayName = 'ButtonStyled';
+
+export const ButtonInnerWrapper = styled.span`
```
david-cho-lerat-sonarsource, line 74: do we want to deal with overflows like we do with dropdown items? Make sure the prefix & suffix are always visible, use '...' on the label?
image

src/components/spinner/Spinner.tsx
jeremy-davis-sonarsource: Yeah, not a fan either. How about:

> Button.tsx
```
const StyledSpinner = styled(Spinner)`
  margin-right: var(--echoes-dimension-size-75);

  background:
    linear-gradient(0deg, var(--button-spinner-color) 50%, transparent 50% 100%) border-box,
    linear-gradient(90deg, var(--button-spinner-color) 25%, transparent 75% 100%) border-box;
`;
```

gregaubert: It would work for this part of the style but not for overriding the margin of the wrapper to 0. Since the classname prop is propagated to the inner spinner.

Another aspect I don't like about this, is that we are moving some style that impact the spinner component in another place unrelated to the spinner itself. I think when looking at the spinner style we should be able to also see all stuff that could impact it's style.

src/components/spinner/Spinner.tsx
```diff
@@ -135,12 +142,22 @@ const SpinnerStyled = styled.span`
...
   border-radius: var(--echoes-border-radius-full);
   vertical-align: text-bottom;
+
+  ${ButtonStyled} & {
+    margin-right: var(--echoes-dimension-size-75);
+
+    background:
+      linear-gradient(0deg, var(--button-spinner-color) 50%, transparent 50% 100%) border-box,
+      linear-gradient(90deg, var(--button-spinner-color) 25%, transparent 75% 100%) border-box;
+  }
```
gregaubert, line 152: Fixing the build with the missing @emotion/babel-plugin, we can now do that! I think that's the cleanest. It keeps all the styles related to the component encapsulated in a single source of truth.

src/components/buttons/ButtonStyles.tsx
```diff
@@ -0,0 +1,174 @@
...
+  display: inline-flex;
```
david-cho-lerat-sonarsource, line 77: here it can be `display: flex;`

david-cho-lerat-sonarsource (approved):
:muscle:

PR #142, opened 2024-07-17
DS-504 Implement ButtonGroup

david-cho-lerat-sonarsource:


src/components/buttons/ButtonGroup.tsx
```diff
@@ -0,0 +1,81 @@
...
+
+import styled from '@emotion/styled';
+import { ReactNode } from 'react';
+
+export enum ButtonGroupAlign {
+  Left = 'left',
+  Right = 'right',
+}
+
+export interface ButtonGroupProps {
```
gregaubert, line 29: Let's add a className

src/components/buttons/ButtonGroup.tsx
```diff
@@ -0,0 +1,81 @@
...
+  align?: ButtonGroupAlign;
+  children: ReactNode;
+  isCombined?: boolean;
+}
+
+export function ButtonGroup({
```
gregaubert, line 35: Let's forward ref and forward the remaining props too

src/components/buttons/ButtonGroup.tsx
```diff
@@ -0,0 +1,81 @@
...
+  flex: 1 0 auto;
+  gap: var(--echoes-dimension-space-100);
+  justify-content: ${({ align }) => (align === ButtonGroupAlign.Left ? 'flex-start' : 'flex-end')};
+  width: 100%;
+`;
+
+const StyledCombinedButtonGroup = styled(StyledButtonGroup)`
+  gap: 0;
+
+  & button {
```
gregaubert, line 59: If you rebase on the last version of my PR you can do `& ${ButtonStyled} {...` instead.

src/components/buttons/ButtonGroup.tsx
david-cho-lerat-sonarsource: where should the ref go? on the container itself?

gregaubert: yeah, we don't have a better alternative anyway

david-cho-lerat-sonarsource: what remaining props could there be in this case?

david-cho-lerat-sonarsource: I'm wondering what the use case would be for the forwarded ref?

gregaubert: Making sure our ButtonGroup is compatible with our other Echoes components that rely on props and ref forwarding.

gregaubert: I don't see a use case right now. I just think it's good practice to make sure all our exported components can have a ref attached to them.
We might have usages where we need to know the size or positioning of the group.

david-cho-lerat-sonarsource: as discussed on Slack, as there's no use case for now and it's just a "structural" component with no inherent behavior, and it's not used through Redux/Mantine, let's keep it simple.

PR #143, opened 2024-07-18
DS-502 Implement ButtonIcon

jeremy-davis-sonarsource:


src/components/buttons/ButtonIcon.tsx
```diff
@@ -0,0 +1,90 @@
...
+import { useButtonClickHandler } from './Button';
+import { BUTTON_VARIETY_STYLES, BUTTONICON_DIMENSIONS, ButtonIconStyled } from './ButtonStyles';
+import { ButtonSize, ButtonVariety, HTMLButtonProps } from './ButtonTypes';
+
+type TooltipProps = ComponentPropsWithoutRef<typeof Tooltip>;
+type TooltipOptions = Omit<TooltipProps, 'children' | 'content' | 'key'>;
+
+export interface ButtonIconProps extends HTMLButtonProps {
+  Icon: React.ForwardRefExoticComponent<IconProps & React.RefAttributes<HTMLSpanElement>>;
+  'aria-label': string;
```
gregaubert, line 34: Should be `ariaLabel`, should we also add a `ariaLabelledBy`?

src/components/buttons/ButtonIcon.tsx
```diff
@@ -0,0 +1,90 @@
...
+import { Tooltip } from '../tooltip';
+import { useButtonClickHandler } from './Button';
+import { BUTTON_VARIETY_STYLES, BUTTONICON_DIMENSIONS, ButtonIconStyled } from './ButtonStyles';
+import { ButtonSize, ButtonVariety, HTMLButtonProps } from './ButtonTypes';
+
+type TooltipProps = ComponentPropsWithoutRef<typeof Tooltip>;
+type TooltipOptions = Omit<TooltipProps, 'children' | 'content' | 'key'>;
+
+export interface ButtonIconProps extends HTMLButtonProps {
+  Icon: React.ForwardRefExoticComponent<IconProps & React.RefAttributes<HTMLSpanElement>>;
```
gregaubert, line 33: Is the typing also compatible with Icon with the fill props ?

src/components/buttons/ButtonIcon.tsx
```diff
@@ -0,0 +1,90 @@
...
+    Icon,
+    hasAutoFocus = false,
+    isDisabled = false,
+    shouldPreventDefault = false,
+    shouldStopPropagation = false,
+    isLoading = false,
+    onClick,
+    size = ButtonSize.Large,
+    variety = ButtonVariety.Neutral,
+    tooltipContent,
```
gregaubert, line 59: Should it be option and default to the ariaLabel value? I think in most case we are gonna repeat the same value as aria-label here.

src/components/buttons/ButtonIcon.tsx
```diff
@@ -0,0 +1,90 @@
...
+import { ComponentPropsWithoutRef, forwardRef, MouseEvent, useMemo } from 'react';
+import { IconProps } from '../icons';
+import { SpinnerOverrideColor } from '../spinner/SpinnerOverrideColor';
+import { Tooltip } from '../tooltip';
+import { useButtonClickHandler } from './Button';
+import { BUTTON_VARIETY_STYLES, BUTTONICON_DIMENSIONS, ButtonIconStyled } from './ButtonStyles';
+import { ButtonSize, ButtonVariety, HTMLButtonProps } from './ButtonTypes';
+
+type TooltipProps = ComponentPropsWithoutRef<typeof Tooltip>;
+type TooltipOptions = Omit<TooltipProps, 'children' | 'content' | 'key'>;
```
gregaubert, line 30: what is 'key' ?

src/components/buttons/ButtonIcon.tsx
```diff
@@ -0,0 +1,90 @@
...
+          }),
+          [variety, size],
+        )}
+        disabled={isDisabled}
+        onClick={handleClick}
+        ref={ref}
+        type={type}>
+        <SpinnerOverrideColor isLoading={isLoading}>
+          <Icon />
+        </SpinnerOverrideColor>
```
gregaubert, line 85: I think we should have two different rendering for when the `isLoading` is set and when it's not.
Otherwise we pollute the DOM with some spinner stuff while there will never be a spinner in some (most?) buttons.

src/components/buttons/ButtonStyles.tsx
```diff
@@ -74,6 +74,13 @@ export const ButtonStyled = styled.button`
 
 ButtonStyled.displayName = 'ButtonStyled';
 
+export const ButtonIconStyled = styled(ButtonStyled)`
+  padding: 0;
```
gregaubert, line 78: Should this be set with the `--button-padding` property directly inside BUTTONICON_DIMENSIONS ?

src/components/buttons/ButtonStyles.tsx
```diff
@@ -172,3 +179,14 @@ export const BUTTON_SIZE_STYLE = {
     '--button-height': 'var(--echoes-sizes-buttons-large)',
   },
 };
+
+export const BUTTONICON_DIMENSIONS = {
```
gregaubert, line 183: Maybe add `_STYLE` suffix to match the other names ?

gregaubert (approved):
It looks great! I just have a few small comments.

stories/ButtonIcon-stories.tsx
```diff
@@ -0,0 +1,52 @@
...
+import { ButtonIcon } from '../src/components/buttons/ButtonIcon';
+
+const meta: Meta<typeof ButtonIcon> = {
+  component: ButtonIcon,
+  title: 'Echoes/ButtonIcon',
+  argTypes: {
+    size: { options: Object.values(ButtonSize) },
+    variety: { options: Object.values(ButtonVariety) },
+  },
+};
```
david-cho-lerat-sonarsource, line 33: maybe add `decorators: [basicWrapperDecorator]`?

stories/ButtonIcon-stories.tsx
```diff
@@ -0,0 +1,52 @@
...
+    'aria-label': 'click the clock',
+    onClick: () => console.log('it is now o`clock'),
+    tooltipContent: 'Click to log the time',
+    tooltipOptions: {
+      align: TooltipAlign.Center,
+      side: TooltipSide.Right,
+    },
+  },
+  render: (args) => {
+    return <ButtonIcon {...args} Icon={IconClock} />;
```
david-cho-lerat-sonarsource, line 50: you could also set the Icon props in args and remove render altogether?

david-cho-lerat-sonarsource (changes requested):
you're missing the export in `buttons/index.ts`

src/components/buttons/ButtonIcon.tsx
jeremy-davis-sonarsource: that would be confusing, I think: Since we expose `HTMLButtonProps`, aria-label is already available, I'm just making it mandatory here.
Also, ariaLabelledBy would be weird: it means there's a label element related to the button. I think in that case the button would be a standard one, not a label + icon button

src/components/buttons/ButtonIcon.tsx
jeremy-davis-sonarsource: I considered it, but we often give context information in the aria-label (e.g. "edit user jean-michel" vs "edit user")

src/components/buttons/ButtonIcon.tsx
gregaubert: 1. HTMLButtonProps doesn't expose aria-label, it only exposes `'form' | 'id' | 'name' | 'role' | 'style' | 'type'`
2. yeah indeed that makes sense for the ariaLabelledby

src/components/buttons/ButtonIcon.tsx
jeremy-davis-sonarsource: it's the ReactKey, added by `ComponentPropsWithoutRef`. I don't think it makes sense to expose that

src/components/buttons/ButtonIcon.tsx
gregaubert: yeah, and I think it's fine to have the same contextual information in the tooltip in most cases, no?
the point is that it avoid having to repeat the value twice when you need the same, it doesn't prevent you to pass a different value for the tooltip if you want to if it's an optional prop with a default value.

src/components/buttons/ButtonIcon.tsx
david-cho-lerat-sonarsource: HTMLButtonProps extend the ARIA props, don't they?

gregaubert: We defined it like this: 
```
export type HTMLButtonProps = Pick<
  ButtonHTMLAttributes<HTMLButtonElement>,
  'form' | 'id' | 'name' | 'role' | 'style' | 'type'
>;
```

gregaubert: `HTMLButtonProps` is not a native type but a custom one we created. Maybe it's too close to real typing name, feel free to rename it to make it more obvious.

david-cho-lerat-sonarsource: interface ButtonHTMLAttributes<T> extends HTMLAttributes<T>
interface HTMLAttributes<T> extends AriaAttributes, DOMAttributes<T>
and AriaAttributes include `"aria-label"?: string | undefined;`

david-cho-lerat-sonarsource: OK my bad, I thought it was the native one

jeremy-davis-sonarsource: Same here, totally got confused

gregaubert: Ok so please rename it to something less confusing

src/components/buttons/ButtonIcon.tsx
gregaubert: Ah I see now that we pass an Icon class, that means we should also have a way to pass the fill icon property

jeremy-davis-sonarsource: I think the fill props are only internal?

gregaubert: No, we have a `isFilled` prop that's public on some icons that implement the `IconFilledProps` instead of the simple `IconProps`

jeremy-davis-sonarsource: It should be fine, since it's extending `IconProps`

gregaubert: Ok, my second point was: right now there is no way to use a filled icon in a ButtonIcon

gregaubert (approved):
Looks good! Well done!

PR #148, opened 2024-07-19
DS-504 Update display value of ButtonGroup to inline-flex

gregaubert:


src/components/buttons/ButtonGroup.tsx
```diff
@@ -41,8 +41,8 @@ export function ButtonGroup({
 }
 
 const StyledButtonGroup = styled.span`
+  display: inline-flex;
```
david-cho-lerat-sonarsource, line 44: NP: why change the order?

david-cho-lerat-sonarsource (approved):
:+1:

src/components/buttons/ButtonGroup.tsx
gregaubert: I usually like to have the display value first and follow with the CSS properties that change it.

gregaubert: like here, the align-items and gap are properties that are tied to display: inline-flex, and change the way the inline flex looks.

david-cho-lerat-sonarsource: ok

PR #150, opened 2024-07-22
DS-516 Select fixes

david-cho-lerat-sonarsource:


src/components/select/SelectCommons.tsx
```diff
@@ -46,7 +47,7 @@ export interface SelectBaseProps {
   labelNotFound?: MantineSelectProps['nothingFound'];
   limit?: MantineSelectProps['limit'];
   name?: MantineSelectProps['name'];
-  optionComponent?: MantineSelectProps['itemComponent'];
+  optionComponent?: (selectItemOptions: SelectOption) => React.JSX.Element | null;
```
gregaubert, line 50: I think we need a generic here, otherwise the typing of the component in the products is gonna be hard when you want to add more stuff on your options.

src/components/select/SelectItemCommons.tsx
```diff
@@ -42,13 +42,19 @@ function withSelectItemWrapper<P extends SelectOption>(
   optionType: SelectOptionType,
 ) {
   const Wrapper = forwardRef<HTMLDivElement, P>((props, ref) => {
-    const { helpText, label, prefix, suffix, value, ...rest } = props;
+    const { disabled, group, helpText, label, prefix, selected, suffix, value, ...rest } = props;
```
gregaubert, line 45: here we want to pass all the options props to the WrappedComponent and options being a generic, I think we need to extract all Mantine props instead of all options props. Mantine props should be fixed on the contrary of options props.

src/components/select/SelectCommons.tsx
jeremy-davis-sonarsource: I think it's fine: SelectOption includes `[key: string]: any;` so it fits any type that has a `value: string`

src/components/select/SelectItemCommons.tsx
jeremy-davis-sonarsource: If I understand properly, I agree!
In other words, we want all the custom props that the consumer passes in the options to be forwarded to the wrapped component, but we don't want the technical props from Mantine to be.

HelpText, Prefix & suffix are already being used, so there's no reason to forward them either.

david-cho-lerat-sonarsource: Right, I'm effectively removing custom props here!
Is there a reference of the Mantine props passed to the `itemComponent` somewhere? [Here](https://v6.mantine.dev/core/select/?t=props) it says `any`..

gregaubert: I don't think we have a reference, we need to check by ourselves what's passed down.

src/components/select/SelectCommons.tsx
gregaubert: Ah yeah nice catch so I think we are good, it's simpler not to have to rely on generic

jeremy-davis-sonarsource (changes requested):
I think the style fix is fine. The prop forwarding is off, though

src/components/select/SelectItemCommons.tsx
```diff
@@ -42,13 +42,30 @@ function withSelectItemWrapper<P extends SelectOption>(
   optionType: SelectOptionType,
 ) {
   const Wrapper = forwardRef<HTMLDivElement, P>((props, ref) => {
-    const { helpText, label, prefix, suffix, value, ...rest } = props;
+    const { helpText, label, prefix, selected, suffix, value, ...mantineAndCustomProps } = props;
```
david-cho-lerat-sonarsource, line 45: I've explicitly extracted `selected` because it's the only one that `SelectItemStatus` needs

src/components/select/SelectItemCommons.tsx
```diff
@@ -42,13 +42,30 @@ function withSelectItemWrapper<P extends SelectOption>(
...
+      onMouseEnter,
+      tabIndex,
+      ...customProps
+    } = mantineAndCustomProps;
+
     return (
-      <SelectItemWrapper ref={ref} {...rest}>
-        <SelectItemStatus optionType={optionType} {...props} />
+      <SelectItemWrapper ref={ref} selected={selected} {...mantineAndCustomProps}>
+        <SelectItemStatus optionType={optionType} selected={selected} />
```
david-cho-lerat-sonarsource, line 61: I figured there's no need to pass `SelectItemStatus` the whole of `props`

src/components/select/SelectItemCommons.tsx
```diff
@@ -42,13 +42,30 @@ function withSelectItemWrapper<P extends SelectOption>(
...
+      onMouseDown,
+      onMouseEnter,
+      tabIndex,
+      ...customProps
+    } = mantineAndCustomProps;
+
     return (
-      <SelectItemWrapper ref={ref} {...rest}>
-        <SelectItemStatus optionType={optionType} {...props} />
+      <SelectItemWrapper ref={ref} selected={selected} {...mantineAndCustomProps}>
```
david-cho-lerat-sonarsource, line 60: `SelectItemWrapper` will receive more than it needs but it's no big deal, it was already the case. It's easier than explicitly extract the Mantine props.

src/components/select/SelectItemCommons.tsx
gregaubert: I think I would only extract our own implementation stuff here: helpText, prefix, suffix

so that all the option fields end up in `mantineAndCustomProps` and you don't have to add them back after in both components

src/components/select/SelectItemCommons.tsx
```diff
@@ -42,13 +42,30 @@ function withSelectItemWrapper<P extends SelectOption>(
...
+    const {
+      'aria-selected': _ariaSelected,
+      className,
+      'data-disabled': _dataDisabled,
+      'data-hovered': _dataHovered,
+      'data-selected': _dataSelected,
+      onMouseDown,
+      onMouseEnter,
+      tabIndex,
+      ...customProps
```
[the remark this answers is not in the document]
gregaubert: You could rename it to `customComponentProps`, to make it more related to the custom component, and not restrict it to the custom fields in the option.

Also it would be nice to add a comment saying something like: `// remove the mantine props from the custom component props` to make it more understandable why all these extracted things are not used later one.

src/components/select/SelectItemCommons.tsx
gregaubert: That's fine, we should not think too much about it, that's the way mantine works if you don't pass a custom component anyway, I think it's good enough if we just remove our own stuff (helpText, prefix, suffix).

Also let's have the `selected` part of `mantineAndCustomProps`

src/components/select/SelectItemCommons.tsx
gregaubert, line 61: just do `selected={props.selected}` here to be able to have `selected` part of the `mantineAndCustomProps`

src/components/select/SelectItemCommons.tsx
```diff
@@ -42,13 +42,30 @@ function withSelectItemWrapper<P extends SelectOption>(
...
     return (
-      <SelectItemWrapper ref={ref} {...rest}>
-        <SelectItemStatus optionType={optionType} {...props} />
+      <SelectItemWrapper ref={ref} selected={selected} {...mantineAndCustomProps}>
+        <SelectItemStatus optionType={optionType} selected={selected} />
         {prefix}
         <SelectItemInner>
-          {WrappedComponent ? <WrappedComponent {...props} /> : <span>{props.label}</span>}
+          {WrappedComponent ? (
+            <WrappedComponent {...({ label, selected, value, ...customProps } as P)} />
```
gregaubert, line 65: We can only pass `...customProps` here if we don't extract `label`, `selected` and `value`. Also, maybe the `as P` will become useless.

jeremy-davis-sonarsource (approved):
Nice work!

src/components/select/SelectItemCommons.tsx
```diff
@@ -42,15 +42,37 @@ function withSelectItemWrapper<P extends SelectOption>(
   optionType: SelectOptionType,
 ) {
   const Wrapper = forwardRef<HTMLDivElement, P>((props, ref) => {
-    const { helpText, label, prefix, suffix, value, ...rest } = props;
+    const { helpText, prefix, suffix, ...mantineAndCustomProps } = props;
+
+    const {
+      // below: we're removing the Mantine props we don't want passed down to the WrappedComponent
```
gregaubert, line 48: nice! 👍🏻

gregaubert (approved):
Great!

PR #152, opened 2024-07-24
DS-299 Add some missing icons from MIUI

gregaubert:


david-cho-lerat-sonarsource (approved):
looking good :+1: 
(hopefully we can get the "org" one too :wink: )

david-cho-lerat-sonarsource:
> looking good 👍 (hopefully we can get the "org" one too 😉 )

haha just commented that as you were adding it; well done!

PR #155, opened 2024-07-24
[NO-JIRA] Prepare for version 0.6-rc1

gregaubert:


src/components/buttons/ButtonGroup.tsx
```diff
@@ -39,6 +39,7 @@ export function ButtonGroup({
 
   return <StyledButtonGroup className={className}>{children}</StyledButtonGroup>;
 }
+ButtonGroup.displayName = 'ButtonGroup';
```
david-cho-lerat-sonarsource, line 42: it's not a `forwardRef`, so its `displayName` should be the name of the function - doesn't it work?

gregaubert: Regardless, we should provide a hardcoded display name like this to all our components; that way, they will render nicely in the react debug tools even after being compiled.

When relying on the function name for the display name, the display name becomes bullshit after the minification of the code.

david-cho-lerat-sonarsource (approved):

PR #160, opened 2024-08-23
Fix compatibility between Selects and Modals

jeremy-davis-sonarsource:


src/components/modals/ModalContext.ts
gregaubert, line 1: Should we make it a bit more generic and call it a `PortalContext`? Ultimately it has nothing to do with the modal, it's just used by it.

src/components/modals/ModalBody.tsx
```diff
@@ -37,13 +38,33 @@ export function ModalBody({ children }: PropsWithChildren<{}>) {
...
+      </ModalContext.Provider>
+
+      {/*
+       * This node is the portal anchor for other overlay types that need to break out of the scrolling content
+       * e.g. Select
+       */}
+      <div
+        ref={(node) => {
+          setPortalRef(node);
+        }}
```
gregaubert, line 65: just pass setPortalRef directly `ref={setPortalRef}`

gregaubert (approved):
Nice!

PR #161, opened 2024-08-23
DS-519 Improvement and fixes for the modal

gregaubert:


src/components/modals/ModalBody.tsx
```diff
@@ -65,11 +70,15 @@ export function ModalBody({ children }: PropsWithChildren<{}>) {
...
+const ModalBodyWrapper = styled.div<Props>`
   position: relative;
   display: flex;
 
-  overflow: hidden;
+  overflow-y: hidden;
+
+  ${({ isLast }) =>
+    isLast &&
+    'border-radius: 0 0 var(--echoes-border-radius-400) var(--echoes-border-radius-400);'}
```
jeremy-davis-sonarsource, line 81: these tokens didn't exist, I replaced them with those that do

jeremy-davis-sonarsource (approved):

src/components/modals/ModalBody.tsx
gregaubert: Thanks Copilot I guess

PR #168, opened 2024-09-09
(DS-321) Changed the structure of the font size and line heights in Echoes.

marciopmoreira6:


design-tokens/tokens/Sonar/base.json
gregaubert, line 1: @marciopmoreira6 I don't think it's expected to have this new file?
It seems to be a duplicate of layer1 base ?

design-tokens/tokens/Themes/dark.json
gregaubert, line 1: Same here these new Themes files are duplicates of layer 2 files

design-tokens/tokens/layer1/base.json
gregaubert, line 1: I don't understand why we have those changes, they don't seems to make sense to me for now. You added a lot of new values that are not used.

design-tokens/tokens/layer1/base.json
```diff
@@ -207,12 +207,42 @@
...
+      },
+      "50": {
+        "value": "2.25rem",
+        "type": "lineHeights"
+      },
+      "60": {
+        "value": "3.125rem",
+        "type": "lineHeights"
+      },
+      "lineHeight": {
```
gregaubert, line 234: This shouldn't be nested, because of that the token name is `...lineHeight.lineHeight.heading`

design-tokens/tokens/layer1/base.json
```diff
@@ -267,6 +297,30 @@
...
+        "type": "fontSizes"
+      },
+      "50": {
+        "value": "1.5rem",
+        "type": "fontSizes"
+      },
+      "60": {
+        "value": "2.125rem",
+        "type": "fontSizes"
+      },
```
gregaubert, line 323: Why these new values?

gregaubert (changes requested):

design-tokens/tokens/layer1/base.json
marciopmoreira6: Yeah, I will do that after we safely remove the old group.

design-tokens/tokens/Sonar/base.json
marciopmoreira6: It was automatically generated. I have no idea why it happened.

design-tokens/tokens/layer1/base.json
marciopmoreira6: The reason for the change in the font size is added in a comment below. 
The old Line-Height system using percentages doesn't make sense because the calculations result in odd numbers and we use a 4px system (pair) for grids and spacings.

design-tokens/tokens/layer1/base.json
gregaubert: What's the logic behind the scale ? I remember for other stuff you tried to have the number 10 or 100 as the base value and then either smaller or bigger. Here the base value seems to be the 20? It will be the most used font-size everywhere right?

design-tokens/tokens/layer1/base.json
gregaubert: so if I understand correctly you want to keep those value for backward compatibility? If that's the case then you should not nest them otherwise the backward compatibility is broken.
So either un-nest them or drop them.

design-tokens/tokens/layer1/base.json
gregaubert: ok make sense, anyway we shouldn't be using these tokens directly much

design-tokens/tokens/layer1/base.json
```diff
@@ -333,7 +387,7 @@
             "value": {
               "fontFamily": "{echoes.fontFamily.sans}",
               "fontWeight": "{echoes.fontWeight.regular}",
-              "lineHeight": "{echoes.lineHeight.body}",
+              "lineHeight": "{echoes.lineHeight.lineHeight.body}",
```
gregaubert, line 390: Let's reference the new lineHeight values no? Same for all the following lineHeights.

And same for the fontsizes.

design-tokens/tokens/layer1/base.json
```diff
@@ -499,8 +550,8 @@
           "value": {
             "fontFamily": "{echoes.fontFamily.sans}",
             "fontWeight": "{echoes.fontWeight.semi-bold}",
-            "lineHeight": "{echoes.lineHeight.body}",
```
marciopmoreira6, line 502: Let's drop Link and Link Large

design-tokens/tokens/layer1/base.json
```diff
@@ -437,18 +488,18 @@
...
-            "fontSize": "{echoes.fontSize.heading.large}",
+            "lineHeight": "{echoes.lineHeight.40}",
+            "fontSize": "{echoes.fontSize.40}",
             "letterSpacing": "{echoes.letterSpacing.decreased}"
           },
           "type": "typography"
         },
-        "xlarge": {
+        "extra large": {
           "value": {
```
marciopmoreira6, line 498: remove the spacing - add a dash

design-tokens/tokens/layer1/base.json
marciopmoreira6: I dropped the values and updated the typography composition with new ones. It shouldn't be a problem now

design-tokens/tokens/layer1/base.json
```diff
@@ -510,12 +561,33 @@
...
-            "lineHeight": "{echoes.lineHeight.body}",
-            "fontSize": "{echoes.fontSize.body.large}",
+            "lineHeight": "{echoes.lineHeight.30}",
+            "fontSize": "{echoes.fontSize.30}",
             "letterSpacing": "{echoes.letterSpacing.default}",
             "textDecoration": "{echoes.textDecoration.underline}"
           },
           "type": "typography"
+        },
+        "helper text": {
```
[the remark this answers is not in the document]
marciopmoreira6: Fixed

design-tokens/tokens/layer1/base.json
```diff
@@ -437,18 +488,18 @@
...
-            "lineHeight": "{echoes.lineHeight.heading}",
-            "fontSize": "{echoes.fontSize.heading.large}",
+            "lineHeight": "{echoes.lineHeight.40}",
+            "fontSize": "{echoes.fontSize.40}",
             "letterSpacing": "{echoes.letterSpacing.decreased}"
           },
           "type": "typography"
         },
-        "xlarge": {
+        "extra large": {
```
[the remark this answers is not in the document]
marciopmoreira6: fixed

design-tokens/tokens/layer1/base.json
```diff
@@ -413,12 +453,23 @@
         }
       },
       "heading": {
+        "extra small": {
```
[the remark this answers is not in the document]
marciopmoreira6: fixed

design-tokens/tokens/layer1/base.json
marciopmoreira6: fixed

design-tokens/tokens/layer1/base.json
marciopmoreira6: fixed

src/generated/design-tokens-base.css
```diff
@@ -93,23 +103,24 @@
...
+  --echoes-typography-heading-small: 700 0.875rem/1.25rem Inter;
+  --echoes-typography-heading-medium: 600 1rem/1.5rem Inter;
+  --echoes-typography-heading-large: 600 1.25rem/1.875rem Inter;
+  --echoes-typography-heading-xlarge: 600 1.5rem/2.25rem Inter;
+  --echoes-typography-code-default: 400 0.875rem/1.25rem 'Ubuntu Mono';
+  --echoes-typography-code-highlight: 700 0.875rem/1.25rem 'Ubuntu Mono';
+  --echoes-typography-code-comment: 400 italic 0.875rem/1.25rem 'Ubuntu Mono';
+  --echoes-typography-others-label: 600 0.875rem/1.25rem Inter;
+  --echoes-typography-others-helper-text: 400 0.75rem/1.125rem Inter;
+  --echoes-typography-display-default: 600 2.125rem/3.125rem Inter;
```
gregaubert, line 123: @jeremy-davis-sonarsource do you want to merge it without updating all the usage of the typography tokens in our components ?

PR #173, opened 2024-09-12
DS-507 Fix Mantine select role and aria-expanded behavior

gregaubert:


jeremy-davis-sonarsource (approved):
:muscle: :muscle: :muscle:

PR #174, opened 2024-09-12
DS-651 Implement Heading component

david-cho-lerat-sonarsource:


stories/Heading-stories.tsx
```diff
@@ -0,0 +1,107 @@
...
+
+export const MediumWithoutMargin: Story = {
+  args: {
+    children: 'A medium heading without margin',
+    hasMarginBottom: false,
+    size: HeadingSize.Medium,
+  },
+  render,
+};
+
```
jeremy-davis-sonarsource, line 72: I'd stick to one story, I don't think it's really useful to have all these different stories for every combination

src/components/typography/Heading.tsx
```diff
@@ -0,0 +1,80 @@
...
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the GNU
+ * Lesser General Public License for more details.
+ *
+ * You should have received a copy of the GNU Lesser General Public License
+ * along with this program; if not, write to the Free Software Foundation,
+ * Inc., 51 Franklin Street, Fifth Floor, Boston, MA  02110-1301, USA.
+ */
+
+import styled from '@emotion/styled';
+import { HeadingProps, Heading as RadixHeading } from '@radix-ui/themes';
```
jeremy-davis-sonarsource, line 22: Not sure about using RadixHeading here. Let's discuss on Monday!

src/components/typography/__tests__/Heading-test.tsx
```diff
@@ -0,0 +1,72 @@
...
+ * Inc., 51 Franklin Street, Fifth Floor, Boston, MA  02110-1301, USA.
+ */
+
+import { screen } from '@testing-library/react';
+import { ComponentProps } from 'react';
+import { render } from '~common/helpers/test-utils';
+import { Heading, HeadingSize } from '../Heading';
+
+it('uses the default size', () => {
+  const container = setupHeading();
```
gregaubert, line 27:
```suggestion
  const { container } = setupHeading();
```

src/components/typography/__tests__/Heading-test.tsx
```diff
@@ -0,0 +1,72 @@
...
+
+  expect(container).toMatchSnapshot();
+});
+
+it.each(['h1', 'h2', 'h3', 'h4', 'h5'])(
+  'renders a Heading as an %s with no bottom margin',
+  async (as: ComponentProps<typeof Heading>['as']) => {
+    const children = `This is a heading rendered as "${as}" with no bottom margin`;
+
+    const container = setupHeading({ as, children });
```
gregaubert, line 37:
```suggestion
    const { container } = setupHeading({ as, children });
```

src/components/typography/Heading.tsx
```diff
@@ -0,0 +1,79 @@
...
+};
+
+type HeadingTag = 'h1' | 'h2' | 'h3' | 'h4' | 'h5';
+
+const defaultSizeByTag: Record<HeadingTag, HeadingSize> = {
+  h1: HeadingSize.ExtraLarge,
+  h2: HeadingSize.Large,
+  h3: HeadingSize.Medium,
+  h4: HeadingSize.Small,
+  h5: HeadingSize.ExtraSmall,
```
gregaubert, line 50: let's add h6 with ExtraSmall too? To follow the documentation in figma...

src/components/typography/Heading.tsx
```diff
@@ -0,0 +1,79 @@
...
+  size?: HeadingSize;
+}
+
+export function Heading({
+  as,
+  children,
+  hasMarginBottom = false,
+  size: sizeProp,
+}: Readonly<PropsWithChildren<Props>>) {
+  const size = sizeProp ?? defaultSizeByTag[as];
```
gregaubert, line 65:
```suggestion
  size = defaultSizeByTag[as],
}: Readonly<PropsWithChildren<Props>>) {
```

src/components/typography/Heading.tsx
```diff
@@ -0,0 +1,79 @@
...
+
+  return <StyledHeading {...{ as, hasMarginBottom, size }}>{children}</StyledHeading>;
```
gregaubert, line 67: why not pass the props like normal?

src/components/typography/Heading.tsx
```diff
@@ -0,0 +1,79 @@
...
+
+const defaultSizeByTag: Record<HeadingTag, HeadingSize> = {
+  h1: HeadingSize.ExtraLarge,
+  h2: HeadingSize.Large,
+  h3: HeadingSize.Medium,
+  h4: HeadingSize.Small,
+  h5: HeadingSize.ExtraSmall,
+};
+
+interface Props {
```
gregaubert, line 53: We are missing the `className` prop

src/components/typography/Heading.tsx
```diff
@@ -0,0 +1,79 @@
...
+  as: HeadingTag;
+  hasMarginBottom?: boolean;
+  size?: HeadingSize;
+}
+
+export function Heading({
+  as,
+  children,
+  hasMarginBottom = false,
+  size: sizeProp,
```
gregaubert, line 63: We should also pass the rest of the props down to the dom and have a forwardRef

src/components/typography/Heading.tsx
david-cho-lerat-sonarsource: I only see five sizes on Figma?

src/components/typography/Heading.tsx
david-cho-lerat-sonarsource: that's what I wrote at first, but `as` is not yet extracted from the props?

src/components/typography/Heading.tsx
david-cho-lerat-sonarsource: it's just more concise

src/components/typography/Heading.tsx
david-cho-lerat-sonarsource: what rest of the props would there be, though, once we add `className`?

david-cho-lerat-sonarsource: do we really need a `forwardRef`? We don't have it systematically on all the other components

david-cho-lerat-sonarsource: OK, added for consistency as discussed

src/components/typography/Heading.tsx
gregaubert: Yes only 5 sizes but seems like it's supposed to support the h6: 
image

src/components/typography/Heading.tsx
gregaubert: mmmh, I guess it's still defined before, because it works fine.

src/components/typography/Heading.tsx
gregaubert: mmmh so you would suggest that we switch to this notation more generally? I'd suggest bringing it to the weekly before we decide to introduce a new coding style on our own.

src/components/typography/Heading.tsx
gregaubert: > what rest of the props would there be, though, once we add `className`?
> 
> * I've added `...rest` anyway

Imagine if you have to pass some `data-something` or `data-testid` for whatever reason. Would be nice that it works out of the box.

src/components/typography/Heading.tsx
david-cho-lerat-sonarsource: I'm not sure how we should interpret this description: it could mean that, where we used either `h5` or `h6` before, we should now use `xsmall` consistently, whatever actual tag ends up being used? Let's ask @marciopmoreira6

src/components/typography/Heading.tsx
david-cho-lerat-sonarsource: oh, OK, I must have messed up when testing this, then. I'll change it

src/components/typography/Heading.tsx
david-cho-lerat-sonarsource: I don't consider this "a new coding style", it's just more compact when the prop name and its value are spelled the same way. This isn't the first time this has been used either.

gregaubert: I'm not so sure. I don't remember reviewing this pattern. If it was commonly used in our codebase I wouldn't be surprised seeing it here.

david-cho-lerat-sonarsource: we have a lot of `{...props}`, which is essentially the same thing, and a lot of objects where we don't have `foo: foo,` but just `foo,` - we even have a rule to make sure this is the case. So I'm not sure what's confusing, here?

src/components/typography/Heading.tsx
david-cho-lerat-sonarsource: but if they want to specifically use `h6` to mean "not `h5`, something smaller," hard-coding a behavior for `h6` that is the same as `h5` won't allow them to do what they want, either. What we really want to know is if, from a design point of view, we *should* have both `h5` and `h6`, as two different things

src/components/typography/Heading.tsx
gregaubert: Ok sure, but I don't think it's nearly the same. We use `{...props}` because there is no other way.
So far, we have always written JSX the JSX way as much as possible, using a syntax that's close to HTML. We are all used to reading and understanding this syntax now and it's a widespread way of doing in the React/JSX community.

To me, this is a new coding style in which we aim for something more concise than our old plain JSX way. Applying this way of doing more generally will change a lot the way our components look (codewise).

It does impact all our devs, so is it something we really want to adopt? And why?

src/components/typography/Heading.tsx
gregaubert: I think he would have added an xxs heading specifically for h6 if we wanted a different style for h6.

src/components/typography/Heading.tsx
jeremy-davis-sonarsource: I'm not a fan of it (i.e. no strong opinion, although I would probably not write it that way), but I think we can keep it as-is, and raise the topic in our weekly.
(this is in echoes, it's not really shared with anyone else)

src/components/typography/Heading.tsx
david-cho-lerat-sonarsource: as per our conversation with @marciopmoreira6, we only allow h1 through h5, and if we ever see h6 in the products we replace it with h5.

src/components/typography/Heading.tsx
david-cho-lerat-sonarsource: I've made this change:

```
-    <StyledHeading {...{ as, ref, size, ...props }} />
+    <StyledHeading as={as} ref={ref} size={size} {...props} />

```
is this OK for you guys?

gregaubert (approved):

PR #176, opened 2024-09-13
DS-520 new 'Display' typographic component

jeremy-davis-sonarsource:


src/components/typography/Display.tsx
```diff
@@ -0,0 +1,47 @@
...
+ * You should have received a copy of the GNU Lesser General Public License
+ * along with this program; if not, write to the Free Software Foundation,
+ * Inc., 51 Franklin Street, Fifth Floor, Boston, MA  02110-1301, USA.
+ */
+
+import styled from '@emotion/styled';
+import { PropsWithChildren } from 'react';
+
+interface Props {
+  as?: 'p' | 'span' | 'div';
```
gregaubert, line 25: `h*` too ?

src/components/typography/Display.tsx
```diff
@@ -0,0 +1,47 @@
...
+  hasMarginBottom?: boolean;
```
gregaubert, line 26: Should we really have this property? When do we want to have margin-bottom for a display? It's not a heading right?

stories/typography/Display-stories.tsx
```diff
@@ -0,0 +1,54 @@
...
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the GNU
+ * Lesser General Public License for more details.
+ *
+ * You should have received a copy of the GNU Lesser General Public License
+ * along with this program; if not, write to the Free Software Foundation,
+ * Inc., 51 Franklin Street, Fifth Floor, Boston, MA  02110-1301, USA.
+ */
+
+/* eslint-disable no-console */
```
gregaubert, line 21: We might not need this line anymore?

david-cho-lerat-sonarsource (approved):
LGTM :+1:

.storybook/preview.tsx
```diff
@@ -32,6 +32,10 @@ const globalStyles = css`
     color: var(--echoes-color-text-default);
     background-color: var(--echoes-color-background-default);
   }
+
+  p {
+    margin: 0;
+  }
```
gregaubert, line 38: I think we shouldn't have that. If we expect the Display component to reset paragraphs margins then it should do so in it's implem, we can't expect the products to have this reset.

src/components/typography/Display.tsx
```diff
@@ -0,0 +1,47 @@
...
+ *
+ * You should have received a copy of the GNU Lesser General Public License
+ * along with this program; if not, write to the Free Software Foundation,
+ * Inc., 51 Franklin Street, Fifth Floor, Boston, MA  02110-1301, USA.
+ */
+
+import styled from '@emotion/styled';
+import { PropsWithChildren } from 'react';
+
+interface Props {
```
gregaubert, line 24: We are missing the className prop

src/components/typography/Display.tsx
```diff
@@ -0,0 +1,47 @@
...
+  as?: 'p' | 'span' | 'div';
+  hasMarginBottom?: boolean;
+}
+
+export function Display({
```
gregaubert, line 29: We should forward ref and rest of the props.

src/components/typography/Display.tsx
jeremy-davis-sonarsource: Good point, I'll ping Marcio

src/components/typography/Display.tsx
jeremy-davis-sonarsource: Specifically not, I believe

src/components/typography/Display.tsx
gregaubert: I also added a comment in figma about it

.storybook/preview.tsx
jeremy-davis-sonarsource: As discussed, it doesn't handle margins at all

PR #178, opened 2024-09-16
DS-653 Implement Label component

david-cho-lerat-sonarsource:


src/components/typography/Label.tsx
```diff
@@ -0,0 +1,38 @@
...
+ * along with this program; if not, write to the Free Software Foundation,
+ * Inc., 51 Franklin Street, Fifth Floor, Boston, MA  02110-1301, USA.
+ */
+
+import styled from '@emotion/styled';
+import { forwardRef, PropsWithChildren } from 'react';
+
+interface Props {
+  className?: string;
+}
```
gregaubert, line 26: Let's add a `as` prop too.

src/components/typography/Label.tsx
```diff
@@ -0,0 +1,38 @@
...
+
+export const Label = forwardRef<HTMLDivElement, Readonly<PropsWithChildren<Props>>>(
+  ({ children, className, ...rest }, ref) => (
+    <StyledLabel {...{ children, className, ref, ...rest }} />
+  ),
+);
```
gregaubert, line 32: I think we don't need this component at all as it's just the StyledLabel itself.

src/components/typography/Label.tsx
```diff
@@ -0,0 +1,38 @@
...
+
+Label.displayName = 'Label';
+
+export const StyledLabel = styled.div`
```
gregaubert, line 36: Let's rename it `Label`, add the `displayName = 'Label'`, restrict the props to your Props interface, and make it a `label` instead of a `div` by default.

src/components/typography/Label.tsx
david-cho-lerat-sonarsource: with which valid values? `div | span`?

gregaubert: at least: `'div' | 'span' | 'label'`

src/components/typography/Label.tsx
```diff
@@ -0,0 +1,32 @@
...
+ * Lesser General Public License for more details.
+ *
+ * You should have received a copy of the GNU Lesser General Public License
+ * along with this program; if not, write to the Free Software Foundation,
+ * Inc., 51 Franklin Street, Fifth Floor, Boston, MA  02110-1301, USA.
+ */
+
+import styled from '@emotion/styled';
+
+interface Props {
```
gregaubert, line 23: We should add an optional `for` prop too, for accessibility

gregaubert (approved):

PR #179, opened 2024-09-16
DS-654 Implement HelperText component

david-cho-lerat-sonarsource:


src/components/typography/HelperText.tsx
gregaubert, line 1: I would say, same comments as for the label component.
Let's add the `as` prop and export the styled component directly instead of wrapping it inside another component.

gregaubert: For the `as` I would say at least `'div' | 'span' | 'p'`

src/components/typography/HelperText.tsx
```diff
@@ -0,0 +1,32 @@
...
+ * along with this program; if not, write to the Free Software Foundation,
+ * Inc., 51 Franklin Street, Fifth Floor, Boston, MA  02110-1301, USA.
+ */
+
+import styled from '@emotion/styled';
+
+interface Props {
+  as?: 'div' | 'p' | 'span';
+  className?: string;
+}
```
gregaubert, line 26: We should allow an `id` prop, so that it can be referenced with `aria-describedby` in the associated input.

gregaubert (approved):

PR #182, opened 2024-09-18
DS-520 Style html tags within Text components

jeremy-davis-sonarsource:


src/components/typography/Text.tsx
```diff
@@ -76,8 +76,22 @@ Text.displayName = 'Text';
...
 
-const StyledText = styled.span<StyledTextProps>`
+const BaseStyles = styled.span`
   max-width: var(--echoes-sizes-typography-max-width-default);
+
+  & ul,
+  & ol {
+    margin-block-start: var(--echoes-dimension-space-100);
+    margin-block-end: 0;
+    padding-inline-start: 28px;
```
jeremy-davis-sonarsource, line 86: This is really specific to the implementation. Not sure we want a token for it

david-cho-lerat-sonarsource (approved):
LGTM :+1:

PR #184, opened 2024-09-18
DS-520 Remove exports of inner styled component of Display and Heading

gregaubert:


jeremy-davis-sonarsource (approved):
Nice catch!

david-cho-lerat-sonarsource (approved):
Thanks!

PR #186, opened 2024-09-19
DS-520 Add strong, b, em, i as possible tags for Text and rename isStrong to isHighlighted

gregaubert:


jeremy-davis-sonarsource (approved):

david-cho-lerat-sonarsource (approved):
Nice :+1:

PR #187, opened 2024-09-19
DS-660 Add tailwind utilities

jeremy-davis-sonarsource:


src/utils/tailwind-utilities-plugin.ts
```diff
@@ -0,0 +1,81 @@
...
+    '.code': {
+      font: 'var(--echoes-typography-code-default)',
+    },
+    '.code-highlight': {
+      font: 'var(--echoes-typography-code-highlight)',
+    },
+    '.code-comment': {
+      font: 'var(--echoes-typography-code-comment)',
+    },
+    '.typo-label': {
```
david-cho-lerat-sonarsource, line 69: NP: why not group all the `.typo-` together?
NP2: and alpha sort them? :P

PR #190, opened 2024-09-20
DS-520 Use Typography components internally

gregaubert:


src/components/dropdown-menu/DropdownMenu.tsx
```diff
@@ -154,19 +159,12 @@ const StyledHeaderLabelAndHelpText = styled.div`
   row-gap: var(---echoes-dimension-space-50);
 `;
 
-const StyledHeaderLabel = styled.div`
-  font: var(--echoes-typography-text-default-semi-bold);
-  overflow: hidden;
-  text-overflow: ellipsis;
-  white-space: nowrap;
+const StyledHeaderLabel = styled(Label)`
+  ${truncate}
```
jeremy-davis-sonarsource, line 163: :clap:

src/components/dropdown-menu/DropdownMenuItemBase.tsx
```diff
@@ -127,7 +127,7 @@ const StyledIconCheck = styled(IconCheck)`
   font-size: var(--echoes-font-size-20);
 `;
 
-const CheckmarkPlaceholder = styled(StyledIconCheck)`
+const CheckmarkPlaceholder = styled(IconCheck)`
   visibility: hidden;
 `;
```
jeremy-davis-sonarsource, line 132: Why this change?

jeremy-davis-sonarsource (approved):
Nice!

src/components/dropdown-menu/DropdownMenuItemBase.tsx
gregaubert: Nice catch, it should be reverted, I was testing if we really need the StyledIconCheck or not and we do.

PR #192, opened 2024-09-23
DS-655 Update Echoes dependencies

gregaubert:


jeremy-davis-sonarsource (approved):
As discussed, remove the snap exclusion and add storybook-static

PR #201, opened 2024-10-11
DS-688 Fix DropdownMenu support of theme overriding

gregaubert:


jeremy-davis-sonarsource (approved):

src/components/dropdown-menu/DropdownMenu.tsx
```diff
@@ -17,12 +17,13 @@
  * along with this program; if not, write to the Free Software Foundation,
  * Inc., 51 Franklin Street, Fifth Floor, Boston, MA  02110-1301, USA.
  */
-
```
david-cho-lerat-sonarsource, line 20: NP: I tend to add a new line below the file header when it's not there, as I believe it's more legible and also to try to have some consistency (some files already have it, some don't). Here, I see all the files in `/dropdown-menu/` do have the new line, so what's the rationale behind removing it here? (maybe it wasn't intentional?)
More generally, I believe we should have a prettier rule to make this consistent at some point.

gregaubert: I added it back, but yeah, we should automate it if we want it to be consistent. I don't want us to waste time discussing blank lines in PRs.

PR #205, opened 2024-10-31
DS-659 Allow elements in RadioOption helpText

david-cho-lerat-sonarsource:


stories/RadioButtonGroup-stories.tsx
gregaubert, line 1: We shouldn't have to replace the stories, a string is a valid ReactNode

gregaubert (approved):

jeremy-davis-sonarsource (approved):
Nice!

stories/RadioButtonGroup-stories.tsx
jeremy-davis-sonarsource: They're just a couple of examples

david-cho-lerat-sonarsource: Yes, some `helpText` are left as `string`, and some use a `ReactNode`, just to reflect that we can.

PR #206, opened 2024-10-31
DS-698 New brand logos

jeremy-davis-sonarsource:
Designs [are here](https://www.figma.com/design/3MMKlRsuGgH5fhU95zrVEv/branch/A1fMWy7ai4FytHunPROcVw/(0.9.0-Echoes)-Foundations?node-id=2856-6578&node-type=symbol&m=dev)


src/components/logos/LogoSonar.tsx
```diff
@@ -0,0 +1,120 @@
...
+  const intl = useIntl();
+
+  return hasText ? (
+    <svg
+      fill="none"
+      height="48"
+      ref={ref}
+      role="img"
+      viewBox="0 0 152 48"
+      width="152"
```
gregaubert, line 37: Could having the height and width hardcoded like this prevent us to use the logos to just fill some available space? or just passing `height="100%" width="100%"` would work fine?

gregaubert (approved):
Looks good!

By any chance did you pass the SVG through some size optimizer like [svgomg](https://svgomg.net/)?

Also, I don't think you can do much, it might be more on the side of @marciopmoreira6. It might be just my screen/computer but it seems like there is a lot of aliasing on the new SQ logos, especially on the colored parts:

image

And even zooming in, the lines don't seem very clean.

src/components/logos/LogoSonar.tsx
jeremy-davis-sonarsource: Hmm, I didn't think we'd have variable sizes on these. I'll check with Marcio

david-cho-lerat-sonarsource:
> Looks good!
> 
> By any chance did you pass the SVG through some size optimizer like [svgomg](https://svgomg.net/)?
> 
> Also, I don't think you can do much, it might be more on the side of @marciopmoreira6. It might be just my screen/computer but it seems like there is a lot of aliasing on the new SQ logos, especially on the colored parts:
> 
> image
> 
> And even zooming in, the lines don't seem very clean.

`yarn dev` didn't start for me because of the `<Svg></svg>` mismatch in `LogoSonar`, but once fixed I have clear-cut images, no aliasing
image

gregaubert (approved):
Code looks good!

I still have a bit of aliasing/bleeding, but that might just be my computer/screen...

david-cho-lerat-sonarsource:
> I still have a bit of aliasing/bleeding, but that might just be my computer/screen...
Yes, I believe it's on your side, it looks fine for me

david-cho-lerat-sonarsource (approved):
LGTM :+1:

src/components/logos/LogoSonar.tsx
```diff
@@ -0,0 +1,117 @@
...
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the GNU
+ * Lesser General Public License for more details.
+ *
+ * You should have received a copy of the GNU Lesser General Public License
+ * along with this program; if not, write to the Free Software Foundation,
+ * Inc., 51 Franklin Street, Fifth Floor, Boston, MA  02110-1301, USA.
+ */
+import { forwardRef } from 'react';
```
david-cho-lerat-sonarsource, line 20: NP: we should update the ESLint rule to have consistent newlines under the headers, like in SQS :stuck_out_tongue_closed_eyes:

PR #208, opened 2024-11-06
DS-699 Add dropdown sub-menus

david-cho-lerat-sonarsource:


src/components/dropdown-menu/DropdownMenu.tsx
```diff
@@ -172,17 +174,6 @@ const StyledHeaderHelpText = styled(HelperText)`
...
-  max-height: var(--radix-dropdown-menu-content-available-height);
-  max-width: var(--echoes-sizes-overlays-max-width-default);
-  min-width: var(--echoes-sizes-overlays-min-width-default);
-  overflow-x: hidden;
-  overflow-y: auto;
-  padding: var(--echoes-dimension-space-50) var(--echoes-dimension-space-0);
-`;
+const StyledDropdownMenuContent = StyledDropdownMenuSubContent.withComponent(
+  radixDropdownMenu.Content,
+);
```
jeremy-davis-sonarsource, line 179: I would have expected the reverse, or maybe even a common base that they both use (e.g. `StyledItem.withComponent`)

jeremy-davis-sonarsource (commented):
The approach looks good, with the minor exception of the shared styling

src/components/dropdown-menu/DropdownMenu.tsx
david-cho-lerat-sonarsource: yes, my concern was that DropdownMenu imports DropdownMenuSubMenu, so I wanted to avoid a circular dependency, and also avoid a new common base

gregaubert: Extracting it to its own file is the way to go I think. We don't want the main component to depend on some "edge case" component styling.

src/components/dropdown-menu/DropdownMenuSubMenu.tsx
```diff
@@ -0,0 +1,112 @@
...
+
+interface DropdownMenuSubProps extends Omit<radixDropdownMenu.DropdownMenuSubProps, 'children'> {
+  className?: string;
+  id?: string;
+  isDisabled?: boolean;
+  isOpen?: boolean;
+  isOpenOnMount?: boolean;
+  items: ReactNode | undefined;
+  triggerLabel: string;
+  triggerSuffix?: ReactNode;
```
gregaubert, line 41: We really want to allow to customize the suffix ?

src/components/dropdown-menu/DropdownMenuSubMenu.tsx
```diff
@@ -0,0 +1,112 @@
...
+import { DropdownMenuItemBase } from './DropdownMenuItemBase';
+
+export type A11yAttrs = Pick<React.AriaAttributes, 'aria-controls'> & {
+  id?: string;
+};
+
+interface DropdownMenuSubProps extends Omit<radixDropdownMenu.DropdownMenuSubProps, 'children'> {
+  className?: string;
+  id?: string;
+  isDisabled?: boolean;
```
gregaubert, line 36: Do we really want to disable a whole submenu? Is it pushed by Marcio? I would expect that you can always see what's inside a submenu, and maybe every options are disabled inside. I don't see the point to disable a whole submenu?

src/components/dropdown-menu/DropdownMenuSubMenu.tsx
```diff
@@ -0,0 +1,112 @@
...
+  isOpen?: boolean;
+  isOpenOnMount?: boolean;
+  items: ReactNode | undefined;
+  triggerLabel: string;
+  triggerSuffix?: ReactNode;
+}
+
+export function DropdownMenuSubMenu({
```
gregaubert, line 44: It seems like we are missing the forwardRef?
It might also explain why it didn't work with the asChild.

src/components/dropdown-menu/DropdownMenuSubMenu.tsx
```diff
@@ -0,0 +1,112 @@
...
+  isDisabled = false,
+  isOpen,
+  isOpenOnMount,
+  items,
+  triggerLabel,
+  triggerSuffix,
+  ...radixProps
+}: Readonly<DropdownMenuSubProps>) {
+  const theme = useContext(ThemeContext);
+  const themeOverrideProp = isDefined(theme) ? { [THEME_DATA_ATTRIBUTE]: theme } : {};
```
gregaubert, line 56: This is not needed since DropdownMenuSubMenu is not rendered in a portal. It's a direct child of a component that rendered in a portal, but that's not an issue since the theme is handled there.

src/components/dropdown-menu/DropdownMenuSubMenu.tsx
```diff
@@ -0,0 +1,112 @@
...
+
+  const a11yAttrs: A11yAttrs = {
+    'aria-controls': `${id}-dropdown-sub`,
+    id: `${id}-trigger-sub`,
+  };
```
gregaubert, line 61: I'm surprised we have to deal with that instead of radix...

src/components/dropdown-menu/DropdownMenuSubMenu.tsx
```diff
@@ -0,0 +1,112 @@
...
+    return trigger;
+  }
+
+  return (
+    <radixDropdownMenu.Sub defaultOpen={isOpenOnMount} open={isOpen}>
+      {trigger}
+
+      <StyledDropdownMenuSubContent
+        {...themeOverrideProp}
+        alignOffset={-4}
```
gregaubert, line 88: We can't really use design tokens for those offset values right?

src/components/dropdown-menu/DropdownMenuSubMenu.tsx
david-cho-lerat-sonarsource: I don't know. I used IconArrowRight as default because it seemed reasonable, but it might not always be the best representation?

src/components/dropdown-menu/DropdownMenuSubMenu.tsx
david-cho-lerat-sonarsource: would we need the forwardRef here, though? This component is never going to be used on its own, only as part of DropdownMenu?

src/components/dropdown-menu/DropdownMenuSubMenu.tsx
david-cho-lerat-sonarsource: I don't know that we need it, I just sort of replicated what was in DropdownMenu :sweat_smile:

gregaubert: Oh wow, we needed to do that manually in DropdownMenu?

src/components/dropdown-menu/DropdownMenuSubMenu.tsx
gregaubert: I would just do it to be safe, it's a public and exported component, I would add the forwardRef.

I think we should have an internal rule for the design system that says: "If it's exported, we forwardRef and props." it's just gonna avoid us future headaches and always having the same discussion.

src/components/dropdown-menu/DropdownMenuSubMenu.tsx
david-cho-lerat-sonarsource: discussed with Marcio, he agrees with you, I'm removing isDisabled

src/components/dropdown-menu/DropdownMenuSubMenu.tsx
david-cho-lerat-sonarsource: apparently, I'm the one who added it, but I'm pretty sure it was by recommendation from one of you guys - probably @jeremy-davis-sonarsource ? Does that ring a bell?

src/components/dropdown-menu/DropdownMenuSubMenu.tsx
```diff
@@ -0,0 +1,83 @@
...
+      isOpen,
+      isOpenOnMount,
+      items,
+      ...radixProps
+    } = props;
+
+    const a11yAttrs: A11yAttrs = {
+      'aria-controls': `${id}-subdropdown`,
+      id: `${id}-subtrigger`,
+    };
```
gregaubert, line 51: From what I can see on radix demo component, these arias fields are handled by radix:
image
image

gregaubert: If we did the same thing in the DropdownMenu base component, then I think we should investigate why we did it there and either drop it or comment on the reason why we have it there.
But radix also attach these props on the main trigger and main dropdown, so I think this is already handled for us and we shouldn't have to worry about it.

src/components/dropdown-menu/DropdownMenuSubMenu.tsx
david-cho-lerat-sonarsource: I probably copied this from the legacy `design-system/` Dropdown - removing it

src/components/dropdown-menu/DropdownMenuSubMenu.tsx
david-cho-lerat-sonarsource: I probably copied this from the legacy `design-system/` Dropdown - removing it

src/components/dropdown-menu/DropdownMenuSubMenu.tsx
david-cho-lerat-sonarsource: Marcio confirmed we should use IconChevronRight and it should not be customizable

gregaubert (approved):
Nice!

PR #209, opened 2024-11-07
DS-685 Use subdued color for disabled form field labels

david-cho-lerat-sonarsource:


gregaubert (approved):
Looks good!

Did you also verify the checkbox? It might need the same changes as the radiobuttongroup.

david-cho-lerat-sonarsource:
> Looks good!
> 
> Did you also verify the checkbox? It might need the same changes as the radiobuttongroup.

yes, we don't have a "checkbox group", and the individual disabled checkbox is ok (the box itself, the label, the help text https://echoes-react.netlify.app/?path=/story/echoes-checkbox--disabled)

PR #210, opened 2024-11-07
Add offset token value used in dropdownMenus

marciopmoreira6:


design-tokens/tokens/layer3/base.json
```diff
@@ -84,6 +84,22 @@
           }
         }
       }
+    },
+    "offset": {
+      "dropdown": {
+        "submenu": {
+          "panel": {
+            "vertical": {
+              "value": "-4px",
```
gregaubert, line 93: As I was saying in slack, I don't think this value should be -4px, that won't be usable on figma...

gregaubert (approved):

david-cho-lerat-sonarsource (approved):

PR #216, opened 2024-11-08
DS-300 Improve RadioButtonGroup a11y

david-cho-lerat-sonarsource:


src/components/radio-button-group/RadioButtonGroup.tsx
```diff
@@ -81,11 +81,18 @@ export const RadioButtonGroup = forwardRef<HTMLDivElement, PropsWithLabels<Props
...
+            <HelperText as="span" id={`${id}-helptext`}>
+              {helpText}
+            </HelperText>
+          )}
         </RadioGroupLabelWrapper>
       )}
 
-      <RadioButtonsWrapper horizontal={alignment === RadioButtonGroupAlignment.Horizontal}>
+      <RadioButtonsWrapper
+        aria-describedby={`${id}-helptext`}
```
gregaubert, line 93: We should add that only if we actually have a helptext, otherwise it's gonna target an id that doesn't exist.

gregaubert (approved):
One small improvement to be done with the helptext, looks good otherwise!

src/components/radio-button-group/RadioButtonGroup.tsx
```diff
@@ -62,16 +62,16 @@ export const RadioButtonGroup = forwardRef<HTMLDivElement, PropsWithLabels<Props
   return (
     <RadioGroupRoot
       {...radixRadioGroupProps}
-      aria-label={ariaLabel}
-      aria-labelledby={ariaLabelledBy}
+      {...(ariaLabel ? { 'aria-label': ariaLabel } : {})}
+      {...(ariaLabelledBy ? { 'aria-labelledby': ariaLabelledBy } : {})}
```
gregaubert, line 66: I don't think that was needed since those are props we receive, they are either undefined or specified, no need to add extra logic around this.

src/components/radio-button-group/RadioButtonGroup.tsx
```diff
@@ -81,11 +81,18 @@ export const RadioButtonGroup = forwardRef<HTMLDivElement, PropsWithLabels<Props
...
+              {helpText}
+            </HelperText>
+          )}
         </RadioGroupLabelWrapper>
       )}
 
-      <RadioButtonsWrapper horizontal={alignment === RadioButtonGroupAlignment.Horizontal}>
+      <RadioButtonsWrapper
+        {...(helpText ? { 'aria-describedby': `${id}-helptext` } : {})}
+        {...(label ? { 'aria-labelledby': `${id}-label` } : {})}
```
gregaubert, line 94: Why going with the complex object spreading? `aria-describedby={helpText? 'id' : undefined}` should be fine, an undefined prop won't appear in the DOM...

PR #243, opened 2025-01-08
ECHOES-531 Implement FormFieldLabel

gregaubert:
[ECHOES-531](https://sonarsource.atlassian.net/browse/ECHOES-531)

Build doesn't pass because of the broken code from the target branch

### DIFF
--- a/src/components/form/FormFieldLabel.tsx
+++ b/src/components/form/FormFieldLabel.tsx
@@ -18,23 +18,43 @@
  * Inc., 51 Franklin Street, Fifth Floor, Boston, MA  02110-1301, USA.
  */
 import styled from '@emotion/styled';
-import { type ReactNode, forwardRef } from 'react';
+import { forwardRef } from 'react';
+import { Label } from '../typography';
 
-import { useFormFieldContext } from './FormFieldContext';
+interface Props {
+  children: JSX.Element | string;
+  isDisabled?: boolean;
+  isRequired?: boolean;
+  htmlFor: string;
+}
 
-export type FormFieldLabelProps = {
-  children: ReactNode;
-  htmlFor?: string;
-};
-
-export const FormFieldLabel = forwardRef<HTMLLabelElement, FormFieldLabelProps>((props, ref) => {
-  const context = useFormFieldContext();
-  const { htmlFor = context.id, ...rest } = props;
-  return <StyledLabel htmlFor={htmlFor} ref={ref} {...rest} />;
+export const FormFieldLabel = forwardRef<HTMLLabelElement, Props>((props, ref) => {
+  const { children, htmlFor, isDisabled = false, isRequired = false, ...rest } = props;
+  return (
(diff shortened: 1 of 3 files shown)


src/components/form/FormFieldLabel.tsx
```diff
@@ -18,23 +18,43 @@
  * Inc., 51 Franklin Street, Fifth Floor, Boston, MA  02110-1301, USA.
  */
 import styled from '@emotion/styled';
-import { type ReactNode, forwardRef } from 'react';
+import { forwardRef } from 'react';
+import { Label } from '../typography';
 
-import { useFormFieldContext } from './FormFieldContext';
+interface Props {
+  children: JSX.Element | string;
```
daniel-nagy, line 25: The global `JSX` namespace is deprecated

image

src/components/form/index.ts
```diff
@@ -17,7 +17,7 @@
  * along with this program; if not, write to the Free Software Foundation,
  * Inc., 51 Franklin Street, Fifth Floor, Boston, MA  02110-1301, USA.
  */
-export { FormField, type FormFieldProps, WithProps, type WithPropsProps } from './FormField';
+export { FormField, WithProps, type FormFieldProps, type WithPropsProps } from './FormField';
 export { FormFieldState } from './FormFieldContext';
-export { FormFieldLabel, type FormFieldLabelProps } from './FormFieldLabel';
+export { FormFieldLabel } from './FormFieldLabel';
```
daniel-nagy, line 22: [question] Are we intentionally not exporting the prop types?

src/components/form/FormFieldLabel.tsx
```diff
@@ -18,23 +18,43 @@
...
-  children: ReactNode;
-  htmlFor?: string;
-};
-
-export const FormFieldLabel = forwardRef<HTMLLabelElement, FormFieldLabelProps>((props, ref) => {
-  const context = useFormFieldContext();
-  const { htmlFor = context.id, ...rest } = props;
-  return <StyledLabel htmlFor={htmlFor} ref={ref} {...rest} />;
+export const FormFieldLabel = forwardRef<HTMLLabelElement, Props>((props, ref) => {
+  const { children, htmlFor, isDisabled = false, isRequired = false, ...rest } = props;
```
daniel-nagy, line 32: [non-blocking] `htmlFor` doesn't need to be desrtructured from the rest of the props

daniel-nagy (approved):
LGTM

src/components/form/FormFieldLabel.tsx
gregaubert: Arf it sucks that it's not reported by our tooling

src/components/form/index.ts
gregaubert: Actually I think we don't even want to export the component either, if we do it will be made public.
I think we usually only export the props if we have a need.

PR #245, opened 2025-01-09
ECHOES-531 Fix border-success and border-info design tokens values

gregaubert:
[ECHOES-531](https://sonarsource.atlassian.net/browse/ECHOES-531)

Part of


gregaubert:
@daniel-nagy --echoes-color-border-success is what we should use for the border of the input text when in success state, I just synched with marcio to fix it and have it as a correct design token instead of a raw value

PR #246, opened 2025-01-09
ECHOES-455 Form Components [WIP]

daniel-nagy:
[ECHOES-455](https://sonarsource.atlassian.net/browse/ECHOES-455)

This is a work in progress but opening a PR for transparency.

### New Components

* CheckboxGroup
* FormField (internal)
* Form
* Form.Header
* Form.Section
* Form.Footer
* TextArea
* TextInput
* ModalForm

### Bug Fixes

* Fixed text overflowing the `MessageCallout` component
* Added `className` prop to `RadioButtonGroup` component.
* Added semi-bold font-weight to text links.

#### Checkbox

- The checkbox will now always have an id. As consequence, the second argument to the `onCheck` callback will never be `undefined`.
- The `aria-disabled` property on the checkbox container was renamed to `data-disabled`.
- The container no longer renders as a label element. Now, the label renders as a label element. See [Wrapping Everything Inside \<label>](https://www.smashingmagazine.com/2023/02/guide-accessible-form-validation/#wrapping-everything-inside-label) for a justification.
- The disabled styles were removed from the label and description.
- The private `HelpText` component was removed and replaced with the `HelperText` component.
- `color: var(--echoes-color-text-default);` was added to the checkbox label

#### Checkbox

- The `ref` was moved from the checkbox container to the checkbox button.
- The `radixProps` was moved from the checkbox container to the checkbox button.

#### Modal & ModalAlert

- The ref was moved from the trigger to the Modal wrapper
- The radixProps was moved from the trigger to the Modal wrapper

#### Design Tokens

- `--echoes-sizes-input-[x]` has been renamed to `--echoes-sizes-form-field-[x]`.

#### Dropdown Menu

* `Root` was removed from the `DropdownMenu` namespace. To migrate, replace `DropdownMenu.Root` with `DropdownMenu`.

#### Radio Button Group

- `labelRadioGroupError` has been removed. To migrate, `labelRadioGroupError` should be renamed to `massageInvalid`, as well as setting the `validation` prop to `"invalid"`.

#### Select & SelectAsync

- `hasError` has been removed. To migrate, `hasError` should be renamed to `validation` and set to `"invalid"`.
- `labelError` has been removed. To migrate, `labelError` should be renamed to `messageInvalid`.
- `size` has been removed. To migrate, remove `size` and use `width` instead.

#### Text

- Small text size now use regular font instead of medium font

[ECHOES-62]: https://sonarsource.atlassian.net/browse/ECHOES-62?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

[ECHOES-455]: https://sonarsource.atlassian.net/browse/ECHOES-455?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


src/common/helpers/useForwardedRef.ts
```diff
@@ -0,0 +1,49 @@
...
+ *
+ * ```typescript
+ * const Input = forwardRef<HTMLInputElement>((props, forwardedRef) => {
+ *   const [ref, setRef] = useForwardedRef(forwardedRef);
+ *   return <input ref={setRef} />
+ * });
+ * ```
+ */
+export function useForwardedRef<T>(forwardedRef: ForwardedRef<T>) {
+  const [ref, setRef] = useState<T | null>(null);
```
gregaubert, line 36: Shouldn't we have a `useRef` here instead?

daniel-nagy: `useRef` will not cause the component to re-render when the ref changes. Meaning it cannot be reliably used as a dependency to a hook. In the select component there is an effect for when the ref changes https://github.com/SonarSource/echoes-react/pull/246/files#diff-d21183601c86fbc1ca9f097f0a4da16299cc8b8c04447581cfb5283cc86bb8acR126

gregaubert: Ok, it makes sense, but then the naming is a bit misleading. I was kind of expecting to get the same type of element as a `useRef`

daniel-nagy: hmm... I could change this to `useRef`, but then I will need to make some changes in the select component to know when the ref changes.

gregaubert: I don't think we should change it, maybe just rename it to something like `useForwardedStateRef` or `useForwardedRefWithState`? Or maybe let's just make it explicit in the jsdoc...

gregaubert (approved):

PR #247, opened 2025-01-10
ECHOES-531 FormField Component

daniel-nagy:
[ECHOES-531](https://sonarsource.atlassian.net/browse/ECHOES-531) Implement `FormField` component.

### Todo

- [x] Revert checkbox component
- [x] Update Select component
- [x] Update RadioGroup component
- [x] Add width prop that restricts form field width
- [x] Update stories
- [x] Add jsdoc comments
- [x] Add tests
- [x] Make sure styling matches Figma
- [x] Make sure semantics are aligned with Figma

### Notes

- This PR reverts the changes to the `Checkbox` component as discussed internally.
- The `FormField` component was refactored a bit to make its usage in various form controls more DRY.
- I created some stories for form fields with "internal" in the name of the story. These will be visible to developers but hopefully labeling them as internal is sufficient.
- Help text has been moved to the bottom of the form field for `Select` and `RadioButtonGroup` components.

#### Design Tokens

- `--echoes-sizes-input-[x]` has been renamed to `--echoes-sizes-form-field-[x]`.

#### Radio Button Group

- `labelRadioGroupError` has been removed. To migrate, `labelRadioGroupError` should be renamed to `massageInvalid`, as well as setting the `validation` prop to `"invalid"`.

#### Select

- `hasError` has been removed. To migrate, `hasError` should be renamed to `validation` and set to `"invalid"`.
- `labelError` has been removed. To migrate, `labelError` should be renamed to `messageInvalid`.
- `size` has been removed. To migrate, remove `size` and use `width` instead.


src/components/form/__tests__/FormField-test.tsx
```diff
@@ -0,0 +1,128 @@
...
+});
+
+it('displays an asterisk next to the label if the form field is required', () => {
+  render(<FormField isRequired label="Label 2" />);
+
+  const label = screen.getByText((_, element) => element?.textContent === 'Label 2*', {
+    selector: 'label',
+  });
+
+  expect(label).toBeVisible();
```
gregaubert, line 38: A simple `screen.getByLabelText('Input Label*');` doesn't work?

src/components/radio-button-group/__tests__/RadioButtonGroup-test.tsx
```diff
@@ -80,13 +81,17 @@ describe('RadioButtonGroup', () => {
...
-    expect(screen.getByTestId('radio-group-label-wrapper')).toHaveTextContent(
-      'Radio group label*Help text',
-    );
+    const description = screen.getByText('Help text');
+    const label = screen.getByText('Radio group label');
+    const radioGroup = screen.getByRole('radiogroup');
+    expect(description).toBeVisible();
+    expect(radioGroup).toHaveAttribute('aria-describedby', description.id);
+    expect(label).toBeVisible();
+    expect(label).toHaveAttribute('for', 'group1');
```
gregaubert, line 94:
```suggestion
    expect(screen.getByLabelText('Radio group label')).toBeInTheDocument();
```

src/components/select/SelectCommons.tsx
```diff
@@ -102,60 +109,71 @@ export const SelectBase = forwardRef<HTMLInputElement, PropsWithLabels<SelectBas
...
-          portalProps: {
-            target: portalContext.portalReference,
-          },
-          withinPortal: isDefined(portalContext.portalReference),
-        }}
-        data={data}
-        data-variant={highlight}
+      <FormField
+        controlId={controlId}
         description={helpText}
```
gregaubert, line 137: The description is not bound with aria-describedBy anymore, same for the error and valid message.

stories/FormField-stories.tsx
```diff
@@ -0,0 +1,158 @@
...
+ */
+import type { Meta, StoryObj } from '@storybook/react';
+import type { ComponentProps } from 'react';
+import { FormField, FormFieldValidation, FormFieldWidth } from '../src/components/form/FormField';
+
+type FormField = typeof FormField;
+
+const meta: Meta<FormField> = {
+  component: FormField,
+  title: 'Echoes/Forms/FormField (internal)',
```
gregaubert, line 28: Maybe we should do another section for internal stuff: `'Internal/FormField'`

stories/FormField-stories.tsx
```diff
@@ -0,0 +1,158 @@
...
+      },
+    },
+    width: {
+      control: { type: 'select' },
+      options: Object.values(FormFieldWidth),
+      table: {
+        defaultValue: { summary: FormFieldWidth.Full },
+      },
+    },
+  },
```
gregaubert, line 44: That's not setup out of the box?

stories/FormField-stories.tsx
```diff
@@ -0,0 +1,158 @@
...
+  parameters: {
+    docs: {
+      toc: true,
+    },
+  },
+};
+
+// eslint-disable-next-line import/no-default-export
```
gregaubert, line 52: I don't see this issue? Maybe you are using the wrong eslint config? All our stories export it like this.

gregaubert (approved):
Nice work overall! I noticed quite a few things to tackle and improve but overall I like the implem! 👍🏻

src/components/checkbox/Checkbox.tsx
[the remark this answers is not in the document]
daniel-nagy: hmm.. I'm not sure that these changes make sense for the checkbox component, since it is not designed to be used directly in a form field. We have this [ticket](https://sonarsource.atlassian.net/browse/ECHOES-542) to align the checkbox API, so maybe we should hold off and implement changes to the checkbox as part of that ticket?

src/components/checkbox/Checkbox.tsx
```diff
@@ -181,3 +176,28 @@ const CheckboxIndicator = styled(RadixCheckbox.Indicator)`
...
+const LabelWrapper = styled.span`
+  display: inline-flex;
+  flex-direction: column;
+  justify-content: center;
+  align-items: flex-start;
+  margin-left: var(--echoes-dimension-space-100);
+`;
+
+const Label = styled.span`
+  font: var(--echoes-typography-others-label-medium);
```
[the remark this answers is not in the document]
daniel-nagy: captured in https://sonarsource.atlassian.net/browse/ECHOES-542

src/components/checkbox/Checkbox.tsx
```diff
@@ -181,3 +176,28 @@ const CheckboxIndicator = styled(RadixCheckbox.Indicator)`
...
+`;
+
+const HelpText = styled.span`
+  font: var(--echoes-typography-others-helper-text);
+  color: var(--echoes-color-text-subdued);
+
+  [aria-disabled='true'] > & {
+    color: var(--echoes-color-text-disabled);
+  }
+`;
```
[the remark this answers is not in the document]
daniel-nagy: captured in https://sonarsource.atlassian.net/browse/ECHOES-542

src/components/checkbox/Checkbox.tsx
```diff
@@ -181,3 +176,28 @@ const CheckboxIndicator = styled(RadixCheckbox.Indicator)`
...
+  align-items: flex-start;
+  margin-left: var(--echoes-dimension-space-100);
+`;
+
+const Label = styled.span`
+  font: var(--echoes-typography-others-label-medium);
+
+  [aria-disabled='true'] > & {
+    color: var(--echoes-color-text-disabled);
+  }
```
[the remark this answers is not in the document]
daniel-nagy: captured in https://sonarsource.atlassian.net/browse/ECHOES-542

src/components/form/FormField.tsx
```diff
@@ -18,101 +18,194 @@
...
 import styled from '@emotion/styled';
-import { type ComponentProps, forwardRef } from 'react';
+import { type ComponentProps, type JSX, forwardRef } from 'react';
+import { FormFieldControl } from './FormFieldControl';
+import { FormFieldLabel } from './FormFieldLabel';
+import { FormFieldMessage } from './FormFieldMessage';
+import { HelperText } from '../typography';
+import { MessageInline, MessageInlineSize, MessageType } from '../messages';
+
+type FormFieldText = JSX.Element | string | false | null;
```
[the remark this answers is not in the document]
daniel-nagy: I'm not saying we shouldn't make this change, but I'm wondering what the reasoning is behind it. Is it to avoid import side effects? If we re-export everything in a public index file then we will have import side-effects anyways, unless the build tooling strips unused exports.

I think of `FormField` as a module, which may export a number of things, and I don't see an advantage to moving types to a separate file. Unless it fixes circular dependencies. Maybe I'm wrong though 🤔

src/components/form/FormField.tsx
```diff
@@ -18,101 +18,194 @@
...
+    });
+
     return (
-      <StyledFormField
-        data-inline={isInline ? '' : undefined}
-        data-placement={controlPlacement}
-        ref={ref}
-        {...props}
-      />
+      <FormFieldStyled data-width={width} ref={ref} {...props}>
```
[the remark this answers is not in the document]
daniel-nagy: I can make that change but this is were I think having a `FormFieldContext` and possibly a `FormContext`, that cascades, would be useful. For example, we could have a `disabled` prop on the `Form` component that disables the entire form. Using the CSS cascade seems more brittle.

Unfortunately, we can't really use a `FormFieldContext` giving the current design. Since the `FormField` is used internally in the form controls.

src/components/form/__tests__/FormField-test.tsx
daniel-nagy: Unfortunately it does not. By default testing-library doesn't match text spanning multiple elements, and the asterisks is wrapped in a span tag.

src/components/radio-button-group/__tests__/RadioButtonGroup-test.tsx
daniel-nagy: These changes expose an accessibility issue with the `RadioGroupButton` component. With this change the test fails with the following error:

> [!CAUTION]
> Found a label with the text of: Radio group label, however the element associated with this label (\<div />) is non-labellable [https://html.spec.whatwg.org/multipage/forms.html#category-label]. If you really need to label a \<div />, you can use aria-label or aria-labelledby instead.

It doesn't like that the label has a `for` attribute pointing to a `div`. It suggests to give the label an `id` and use `aria-labelledby` on the div, but that creates some inconsistency with the other form controls. The other alternative is to make the RadioButtonGroup a [form associated custom element](https://html.spec.whatwg.org/multipage/forms.html#form-associated-element), which might be the best solution, in terms of accessibility, but a larger change.

stories/FormField-stories.tsx
daniel-nagy: If I just set them as args, e.g.

```typescript
args: {
  ...,
  validation: FormFieldValidation.None,
  width: FormFieldWidth.Full,
}
```

Then they appear as text controls

image

I think we may want to invest more time in general in our storybook documentation. Admittedly I'm not super familiar with it. I was disappointed that storybook does not use my jsdoc comments for documentation.

src/components/form/FormFieldLabel.tsx
```diff
@@ -18,46 +18,66 @@
...
   isDisabled?: boolean;
+  /**
+   * When true, will display an asterisk to indicate that the field is required.
+   */
   isRequired?: boolean;
-  htmlFor: string;
+  /**
+   * The ID of the form control that this label is associated with.
+   */
+  htmlFor?: string;
```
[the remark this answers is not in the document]
daniel-nagy: To do this I would need to make the `controlId` required on the `FormField`. However, there is an issue with the `RadioButtonGroup` component because the root is not a "labelable" element. See https://github.com/SonarSource/echoes-react/pull/247#discussion_r1927481936.

src/components/form/FormField.tsx
```diff
@@ -18,101 +18,194 @@
...
+        </MessageInline>
+      );
+    case Boolean(messageValid) && validation === FormFieldValidation.Valid:
+      return (
+        <MessageInline size={MessageInlineSize.Small} type={MessageType.Success}>
+          {messageValid}
+        </MessageInline>
+      );
+    case Boolean(description):
+      return <HelperText id={descriptionId}>{description}</HelperText>;
```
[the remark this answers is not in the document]
daniel-nagy: I added a separate `validationMessageId` prop for validation messages.

src/components/select/SelectCommons.tsx
daniel-nagy: I added `aria-describedby` to the select component but mantine select doesn't seem to be forwarding it to the input component.

src/components/checkbox/Checkbox.tsx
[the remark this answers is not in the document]
gregaubert: As you said, let's do those in the other ticket.

src/components/form/FormField.tsx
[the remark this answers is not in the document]
gregaubert: I think the main reason is consistency. So far, we have kept types that are re-used or exported in separate Types files. This keeps the FormField file focused on the FormField component implementation, not overloading it with plenty of other stuff. It's probably not the case here, but it can help avoid some circular dependencies.

src/components/form/FormField.tsx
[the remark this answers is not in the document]
gregaubert: Yeah I think a FormFieldContext would be a bit overkill. But it's true the CSS cascade approach I was suggesting is a bit brittle, let's not go this route.
I'll just handle the data-disabled in TextInput directly.

src/components/form/FormFieldLabel.tsx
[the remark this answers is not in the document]
gregaubert: Mmmh, ok, let's forget about it. Anyway, FormFieldLabel is an internal component; it's not like we need type to help users of the lib avoid mistakes.

src/components/form/__tests__/FormField-test.tsx
gregaubert: Or does it not work because we are not passing any children? So there is actually nothing being labeled?
I'm saying this because I'm using `screen.getByLabelText('Input Label*')` in a TextInput test and it's able to match it somehow.

src/components/radio-button-group/__tests__/RadioButtonGroup-test.tsx
gregaubert: Arf ok.
So looking at Deque documentation they indeed do it with aria-labelledBy and the radiogroup role (that I think we are also missing): https://dequeuniversity.com/library/aria/radio-and-radio-group#:~:text=Create%20a%20or%20%3C,label%22%20and%20a%20unique%20ID.

```html
<div class="deque-radio-group"  aria-labelledby="inspire">
  <div class="deque-radio-group-label" id="inspire" >
    Which one of these scientists most inspires you?</div>
  <div class="radioGroup">
    <span id="Whichoneofthesescientistsmostinspiresyou_0" class="deque-radio" aria-labelledby="curie"></span>
    <span id="curie" >Marie Curie</span>

    <span id="Whichoneofthesescientistsmostinspiresyou_1" class="deque-radio" aria-labelledby="goodall"></span>
    <span id="goodall" >Jane Goodall</span>

    <span id="Whichoneofthesescientistsmostinspiresyou_2" class="deque-radio" aria-labelledby="franklin"></span>
    <span id="franklin" >Rosalind Franklin</span>

    <span id="Whichoneofthesescientistsmostinspiresyou_3" class="deque-radio" aria-labelledby="lamarr"></span>
    <span id="lamarr" >Hedy Lamarr</span>
  </div>
</div>
```

src/components/select/SelectCommons.tsx
gregaubert: Arf maybe that means we need to keep using the `description` prop of mantine for that and just re-use the style of FormFieldMessage?

stories/FormField-stories.tsx
gregaubert: Ah that's super weird, I also have similar typing in TextInput-story and it shows this:
image

But I'm passing the args as string instead of using the enum... That's only difference I noticed.

src/components/text-input/TextInput.tsx
```diff
@@ -75,6 +75,14 @@ export const TextField = forwardRef<HTMLInputElement, PropsWithLabels<Props>>((p
...
   const controlId = id ?? defaultId;
   const descriptionId = `${controlId}-description`;
+  const validationMessageId = `${controlId}-validation-message`;
+
+  const describedBy = [
+    Boolean(messageValid || messageInvalid) && validationMessageId,
+    Boolean(helpText) && descriptionId,
+  ]
+    .filter((id) => id)
+    .join(' ');
```
gregaubert, line 85: Should we extract it to a hook reused in each components? We could also put that piece in the hook: `describedBy !== '' ? describedBy : undefined`

gregaubert (approved):
It's really looking good!
Seems like we just need to figure out the issues with Select and RadioGroup now...
Also I'll squash my PR into your to stop having to rebase it

src/components/form/FormField.tsx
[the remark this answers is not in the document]
daniel-nagy: I think I would approach it by limiting the size of the module, not by what's in it. At a little over 200 lines, with plenty of jsdoc comments, I would say this is module is fine. If a module gets too big, maybe approaching 1,000 lines, then I would break it down into smaller modules, but I wouldn't necessarily partition by types vs not types.

src/components/form/__tests__/FormField-test.tsx
daniel-nagy: Ah yes, you're right. When it is associated with a labelable element it works!

src/components/text-input/TextInput.tsx
daniel-nagy: I'll see if I can come up with a hook API that makes sense. This is why I wanted to use a FormField context, so all the logic could be in the form field and the controls could just do something like:

```typescript
const { controlId, describedBy, validation } = useContext(FormFeildContext);
```

stories/FormField-stories.tsx
daniel-nagy: hmm... Yeah storybook seem pretty inconsistent in how it generates controls. Maybe its work updating to see if it has improved?

src/components/form/FormField.tsx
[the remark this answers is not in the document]
gregaubert: Fine, let's keep it like this

src/components/select/SelectCommons.tsx
daniel-nagy: Then we would have to do that with the validation messages as well... I wonder if I can just get a ref to the underlaying input and add the aria attributes imperatively as a workaround?

gregaubert: True, that sounds like a good workaround if you can make it work

daniel-nagy: Seems to be working https://github.com/SonarSource/echoes-react/pull/247/commits/e357063e5bd80d4ba083f212006a7cc443adb967

gregaubert: Nice!

gregaubert: It might be worth extracting the `setRef` helper to a common folder, I believe we might need to get refs internally in more component in the future, it would also probably be nice to make it a bit more explicit in the name that it's keeping a ref for ourself and still setting the one we receive from the parent.

src/components/radio-button-group/__tests__/RadioButtonGroup-test.tsx
daniel-nagy: See https://github.com/SonarSource/echoes-react/pull/247/commits/609fa68d72593fbf4f7f23a6ca1326dc8dfb5c30

gregaubert: Nice!

src/components/select/SelectCommons.tsx
daniel-nagy: See https://github.com/SonarSource/echoes-react/pull/247/commits/e155bb912d25cff943b20fa68198814ec7728e93

src/components/text-input/__tests__/TextInput-test.tsx
```diff
@@ -94,7 +94,7 @@ it('should display an valid message when validation is valid', () => {
   const commonProps = { helpText: 'Helping', label: 'Label', messageValid: 'Youhou!' };
   const { rerender } = render(<TextInput {...commonProps} validation="none" />);
 
-  expect(screen.getByRole('textbox', { description: 'Helping' })).toBeInTheDocument();
+  expect(screen.getByRole('textbox', { description: ' Helping' })).toBeInTheDocument();
```
daniel-nagy, line 97: The text input is described by both the validation message and the description, it seems like the empty validation message is adding a space before the description. I believe this is harmless, but I could be wrong.

stories/FormField-stories.tsx
daniel-nagy: Moving the component definition to the top of the file it fixed it 🤷‍♂️

PR #249, opened 2025-01-24
ECHOES-542 Align Checkbox

daniel-nagy:
[ECHOES-542](https://sonarsource.atlassian.net/browse/ECHOES-542)

#### Notes

The [Figma page](https://www.figma.com/design/wYk5yGLz6MYC2hdu70IGql/(0.13.0)-Form-Fields?node-id=80-200) for checkboxes has not been updated to remove the disabled state from the label and description.

#### Changes

- The checkbox will now always have an id. As consequence, the second argument to the `onCheck` callback will never be `undefined`.
- The `aria-disabled` property on the checkbox container was renamed to `data-disabled`.
- The container no longer renders as a label element. Now, the label renders as a label element. See [Wrapping Everything Inside \<label>](https://www.smashingmagazine.com/2023/02/guide-accessible-form-validation/#wrapping-everything-inside-label) for a justification.
- The disabled styles were removed from the label and description.
- The private `HelpText` component was removed and replaced with the `HelperText` component.
- `color: var(--echoes-color-text-default);` was added to the checkbox label

#### Breaking Changes

- The `ref` was moved from the checkbox container to the checkbox button.
- The `radixProps` was moved from the checkbox container to the checkbox button.

[ECHOES-542]: https://sonarsource.atlassian.net/browse/ECHOES-542?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/stories/Checkbox-stories.tsx
+++ b/stories/Checkbox-stories.tsx
@@ -19,8 +19,8 @@
  */
 
 import type { Meta, StoryObj } from '@storybook/react';
-import { ComponentProps, useEffect, useState } from 'react';
-import { Checkbox } from '../src';
+import { ComponentProps, forwardRef, useEffect, useState } from 'react';
+import { Checkbox, Tooltip } from '../src';
 
 const meta: Meta<typeof Checkbox> = {
   component: Checkbox,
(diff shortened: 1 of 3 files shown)


daniel-nagy:
I believe the build is failing due to an internal error in Cirrus CI.

gregaubert (approved):
The changes look great!

I pinged Marcio to update the Figma too.

I also tried adding this story to see where the tooltip would show up with the move of the ref:
```
export const WithTooltip: Story = {
  args: {
    checked: false,
    label: "I'm a checkbox",
    helpText: 'I have some help text',
    onCheck: () => {},
  },
  render: (args) => (
    <Tooltip content="text">
      <CheckboxState initialValue={args.checked} {...args} />
    </Tooltip>
  ),
};
```
And it turns out the tooltip doesn't show up at all. But it doesn't seem like a new issue, trying the same story on master branch also doesn't work. So I think it's fine to either proceed with the current and create a ticket or if the fix is obvious to you (it's not to me) to fix it there before merging.

daniel-nagy:
> The changes look great!
> 
> I pinged Marcio to update the Figma too.
> 
> I also tried adding this story to see where the tooltip would show up with the move of the ref:
> 
> ```
> export const WithTooltip: Story = {
>   args: {
>     checked: false,
>     label: "I'm a checkbox",
>     helpText: 'I have some help text',
>     onCheck: () => {},
>   },
>   render: (args) => (
>     <Tooltip content="text">
>       <CheckboxState initialValue={args.checked} {...args} />
>     </Tooltip>
>   ),
> };
> ```
> 
> And it turns out the tooltip doesn't show up at all. But it doesn't seem like a new issue, trying the same story on master branch also doesn't work. So I think it's fine to either proceed with the current and create a ticket or if the fix is obvious to you (it's not to me) to fix it there before merging.

The tooltip is not working because the `CheckboxState` component in the story does not use `forwardRef`. When I forward the ref in `CheckboxState` it works. However, the tooltip now only displays if you hover the checkbox control, it does not display if you hover the checkbox label or description. Which is different from how it behaves currently.

**Current**

image

**New**

image

To get the current behavior, the checkbox can be wrapped in a div, e.g.

```tsx
<Tooltip content="test">
  <div>
    <CheckboxState initialValue={args.checked} {...args} />
  </div>
</Tooltip>
```

Do you have any opinion about whether we should forward the ref to the checkbox root or the container? My thought is if someone wants a reference to the checkbox, they want the checkbox root so they can inspect/manipulate its state.

stories/Checkbox-stories.tsx
```diff
@@ -100,19 +100,41 @@ export const LoadingState: Story = {
...
-    <Checkbox
-      {...(checkboxProps as ComponentProps<typeof Checkbox>)}
-      checked={checked}
-      onCheck={setChecked}
-    />
-  );
-}
+};
+
+const CheckboxState = forwardRef<HTMLButtonElement, CheckboxStateProps>(
```
gregaubert, line 125: Good catch!

stories/Checkbox-stories.tsx
```diff
@@ -100,19 +100,41 @@ export const LoadingState: Story = {
...
+    helpText: 'I have some help text',
+    onCheck: () => {},
+  },
+  render: (args) => (
+    <div style={{ display: 'flex', alignItems: 'center' }}>
+      <Tooltip content="test">
+        <div>
+          <CheckboxState initialValue={args.checked} {...args} />
+        </div>
+      </Tooltip>
```
gregaubert, line 116: I think we should remove the wrapping divs, and just target the checkbox with the tooltip.
The tooltip trigger should always be an interactive element, so it makes sense to target the checkbox itself.

PR #251, opened 2025-01-28
ECHOES-376 ECHOES-560 Components hardening

gregaubert:
This PR fixes the two small tickets we had in the sprint about Radio button and MessageCallout.


daniel-nagy (approved):
LGTM! Do we want to target the **form-components** branch or **main**?

gregaubert:
I think targeting form-components is fine, since we touched radiogroup, I don't won't to end up with useless conflicts to solve.

PR #252, opened 2025-01-28
ECHOES-543 add CheckboxGroup component

daniel-nagy:
[ECHOES-543](https://sonarsource.atlassian.net/browse/ECHOES-543)

Creates a `CheckboxGroup` component, which allows grouping multiple checkboxes with a label and description.

**Todo**

- [x] Add tests
- [x] Add jsdoc comments
- [x] Add stories
- [x] Add horizontal styles
- [x] Make sure semantics match Figma
- [x] Make sure styles match Figma

[ECHOES-543]: https://sonarsource.atlassian.net/browse/ECHOES-543?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


src/components/checkbox-group/CheckboxGroup.tsx
```diff
@@ -0,0 +1,282 @@
...
+        width={width}>
+        <CheckboxGroupRoot // NOSONAR
+          aria-describedby={describedBy}
+          aria-label={ariaLabel}
+          aria-labelledby={ariaLabelledBy ?? labelId}
+          data-direction={direction}
+          data-invalid={validation === FormFieldValidation.Invalid}
+          id={controlId}
+          ref={ref}
+          role="group">
```
daniel-nagy, line 129: I have a `// NOSONAR` here because it is complaining about the `group` role.

src/components/checkbox/Checkbox.tsx
```diff
@@ -75,8 +78,9 @@ export const Checkbox = forwardRef<HTMLButtonElement, PropsWithLabels<Props>>((p
         <Spinner isLoading={isLoading}>
           <CheckboxRoot
             aria-disabled={isDisabled}
+            aria-invalid={hasError}
             aria-label={ariaLabel ?? title}
-            aria-labelledby={ariaLabelledBy}
+            aria-labelledby={ariaLabelledBy ?? labelId}
```
gregaubert, line 83: Isn't adding the `labelId` here duplicating the `htmlFor` we have on the `Label` component?

stories/CheckboxGroup-stories.tsx
```diff
@@ -0,0 +1,167 @@
...
+ */
+import type { Meta, StoryObj } from '@storybook/react';
+import { forwardRef, useEffect, useRef, useState } from 'react';
+import {
+  type CheckboxGroupProps,
+  CheckboxGroup,
+} from '../src/components/checkbox-group/CheckboxGroup';
+import { toTextControlArgTypes } from './helpers/arg-types';
+
+type CheckboxGroup = typeof CheckboxGroup;
```
gregaubert, line 28: Using the same name here is disturbing for me. But I guess TypeScript is smart enough to make the difference between the type definition and the actual Component based on the context where it's used.

src/components/checkbox-group/CheckboxGroup.tsx
gregaubert: Let's remove it, we prefer managing these directly in SQ UI by "accepting" the issue and eventually explaining why, instead of in the code.

gregaubert (changes requested):
Nice work overall! Well done!

IMO there are just a few more fixes to be done before merging!

src/components/checkbox/Checkbox.tsx
daniel-nagy: Yeah, but without it axe is complaining that there is no discernible text for the button, even though the button has the `checkbox` role, and a button is a labelable element.

```
"Buttons must have discernible text (button-name)"

Fix any of the following:
  Element does not have inner text that is visible to screen readers
  aria-label attribute does not exist or is empty
  aria-labelledby attribute does not exist, references elements that do not exist or references elements that are empty
  Element has no title attribute
  Element's default semantics were not overridden with role="none" or role="presentation"
```

Do you think we should just get rid of the `for` attribute then?

stories/CheckboxGroup-stories.tsx
daniel-nagy: Yeah, that's correct.

src/components/checkbox-group/CheckboxGroup.tsx
daniel-nagy: I did this but SonarQube still complains about it in IDE

image

src/components/checkbox/Checkbox.tsx
gregaubert: Ok so let's just get rid of the "for" and keep only the labelledBY

src/components/checkbox-group/CheckboxGroup.tsx
gregaubert: Mmmh if SonarQube keeps complaining in IDE after accepting the issue in the UI it's probably because your Sonarlint isn't in connected mode. Lets try to fix that instead of using nosonar

src/components/radio-button-group/RadioButtonGroup.tsx
```diff
@@ -31,17 +32,12 @@ import {
...
-  Horizontal = 'horizontal',
-}
-
 interface Props extends ValidationProps {
   onChange?: (value: string) => void;
   options: RadioOption[];
 
   // Group Props
-  alignment?: RadioButtonGroupAlignment;
+  alignment?: GroupAlignment;
```
gregaubert, line 40: Let's use template literal?

gregaubert (approved):
That last commit message 😂 
Looks great!

src/components/checkbox/Checkbox.tsx
daniel-nagy: If I remove the `for` attribute, then clicking the label does not check the checkbox ☹️

daniel-nagy: Honestly, I think axe is probably wrong in this case

daniel-nagy: I think this was fixed in axe-core [`v4.10.1`](https://github.com/dequelabs/axe-core/issues/3696). Unfortunately, jest-axe `v9.0.0` has a dependency on axe-core `v4.9.1`, and v9 is the latest version.

We can try using a resolution to use a new version of axe-core.

daniel-nagy: If I add a resolution for `"axe-core": "4.10.2"` it fixes it.

daniel-nagy: I can submit an upstream PR to bump axe-core in jest-axe and maybe they will merge it quickly

PR #257, opened 2025-01-31
ECHOES-568 Migration Fixes

daniel-nagy:
[ECHOES-568](https://sonarsource.atlassian.net/browse/ECHOES-568)

There are a bunch of tests failing in the product because Mantine is adding the same label to the input element and option dropdown wrapper element.

[ECHOES-568]: https://sonarsource.atlassian.net/browse/ECHOES-568?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/components/select/SelectCommons.tsx
+++ b/src/components/select/SelectCommons.tsx
@@ -173,6 +173,15 @@ export const SelectBase = forwardRef<HTMLInputElement, PropsWithLabels<SelectBas
           error={validation === FormFieldValidation.Invalid}
           filter={optionsFilter}
           id={controlId}
+          label={label}
+          labelProps={{
+            // We no longer use Mantine's InputLabel component. However, if we
+            // do not pass a `label` prop to the Select component, Mantine will
+            // render an `aria-label` on the OptionsDropdown component. This
+            // causes the input and the dropdown to have the same label, which
+            // is problematic for accessibility.
+            labelElement: Null,
+          }}
           leftSection={valueIcon}
           nothingFoundMessage={labelNotFound}
           onDropdownOpen={onOpen}
@@ -193,6 +202,10 @@ export const SelectBase = forwardRef<HTMLInputElement, PropsWithLabels<SelectBas
     );
   },
 );
+
+function Null() {
+  return null;
+}
 
 SelectBase.displayName = 'SelectBase';
 


src/components/select/SelectCommons.tsx
```diff
@@ -173,6 +173,15 @@ export const SelectBase = forwardRef<HTMLInputElement, PropsWithLabels<SelectBas
...
           id={controlId}
+          label={label}
+          labelProps={{
+            // We no longer use Mantine's InputLabel component. However, if we
+            // do not pass a `label` prop to the Select component, Mantine will
+            // render an `aria-label` on the OptionsDropdown component. This
+            // causes the input and the dropdown to have the same label, which
+            // is problematic for accessibility.
+            labelElement: Null,
+          }}
```
gregaubert, line 184: Looking at the Dom, I can see that the dropdown as a `aria-describedBy` that points to a nonexisting label id:
image

Should we actually add that id to our label? From what I can see on our previous implem this `aria-labelledBy` was correctly pointing to the mantine label, see:
image

gregaubert: It's probably inferring it from the parent element id we are passing

daniel-nagy: In [this example](https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Roles/combobox_role#examples) on MDN, the input and dropdown have different labels. The input as an `aria-controls` that points to the `id` of the drop-down menu.

However, on https://echoes-react.netlify.app/?path=/story/echoes-select-select--default, the input's `aria-controls` does not match the `id` of the drop-down element, but in your screenshot it does? 🤔

Screenshot 2025-02-03 at 9 13 43 AM

What I can take away from this:

1. The input should have an `aria-controls` that points to the `id` of the drop-down menu.
2. The input and drop-down should have different labels.

Neither of these seem correct in Mantine on my end.

daniel-nagy: The `aria-controls` changes once you open the drop-down menu. But it is incorrect until the drop-dowm menu opens.

daniel-nagy: > Ok, so given what you are saying, we want to keep the labelElement: Null to avoid having duplicate labels.

Yeah, I'm not seeing any way to provide a different label for the drop-down menu element in the source code. In that case do you think this change is acceptable?

gregaubert: Yeah I think it's acceptable

gregaubert (approved):

PR #266, opened 2025-02-05
ECHOES-584 Change helper-text typography design token to regular font-weight

gregaubert:
ECHOES-584

This change is following the change that we did for small text

### DIFF
--- a/design-tokens/tokens/layer1/base.json
+++ b/design-tokens/tokens/layer1/base.json
@@ -529,7 +529,7 @@
         "helper-text": {
           "value": {
             "fontFamily": "{echoes.fontFamily.sans}",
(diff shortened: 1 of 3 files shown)


daniel-nagy (approved):
LGTM!

as an aside, I noticed the text of the bread crumbs component changed with the Echoes update, it is now bolder. Is that expected?

image

gregaubert:
> LGTM!
> 
> as an aside, I noticed the text of the bread crumbs component changed with the Echoes update, it is now bolder. Is that expected?
> 
> image

It should actually be the reverse, we changed a lot of text from `medium` to `regular`. So there is something weird to check there!

gregaubert:
Ah no ok it's normal actually. The Links font-weight changed to be bolder, and these are links...

So it's expected!

PR #267, opened 2025-02-05
ECHOES-585 Make modal title optional

daniel-nagy:
[ECHOES-585](https://sonarsource.atlassian.net/browse/ECHOES-585)

Based on https://github.com/SonarSource/echoes-react/commit/4b14eaa12d7173c3886f66bb5d06b866998f8f77#r152181382

### DIFF
--- a/src/components/modals/ModalTypes.ts
+++ b/src/components/modals/ModalTypes.ts
@@ -19,7 +19,7 @@
  */
 
 import { ReactNode } from 'react';
-import { TextNode } from '~types/utils';
+import { TextNodeOptional } from '~types/utils';
 
 export enum ModalSize {
   Default = 'default',
@@ -39,7 +39,7 @@ interface CommonProps {
   content?: ReactNode;
   isDefaultOpen?: boolean;
   /** An accessible title to be announced when the dialog is opened. */
-  title?: TextNode;
+  title?: TextNodeOptional;
 }
 
 interface UncontrolledProps extends CommonProps {


gregaubert:
I would think that instead, we should make the title mandatory for modals...

Let's see if we can get a bit more input on this: https://sonarsource.slack.com/archives/C06HHK3KUCV/p1738772068811179

gregaubert:
Alright so Marcio confirmed that Modals without a title are possible.
So let's make it fully optional. That means we need to make sure it looks ok when no title and that we don't have an empty <h2> at the top of the modal when the title isn't provided.

daniel-nagy:
> Alright so Marcio confirmed that Modals without a title are possible.
> So let's make it fully optional. That means we need to make sure it looks ok when no title and that we don't have an empty `<h2>` at the top of the modal when the title isn't provided.

Given that it is already allowed to render a modal without a title, and this is just fixing a type error that we introduced, do you think we should address this now or create a new ticket to fix this later?

gregaubert:
> > Alright so Marcio confirmed that Modals without a title are possible.
> > So let's make it fully optional. That means we need to make sure it looks ok when no title and that we don't have an empty `<h2>` at the top of the modal when the title isn't provided.
> 
> Given that it is already allowed to render a modal without a title, and this is just fixing a type error that we introduced, do you think we should address this now or create a new ticket to fix this later?

Yeah ok let's create a new ticket to address this later!

gregaubert:
I'll merge this already

daniel-nagy:
Ticket created https://sonarsource.atlassian.net/browse/ECHOES-586

PR #270, opened 2025-02-07
ECHOES-591 Add destructive ModalForm use case

gregaubert:
[ECHOES-591](https://sonarsource.atlassian.net/browse/ECHOES-591)

Part of [ECHOES-455](https://sonarsource.atlassian.net/browse/ECHOES-455)

[ECHOES-455]: https://sonarsource.atlassian.net/browse/ECHOES-455?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

[ECHOES-591]: https://sonarsource.atlassian.net/browse/ECHOES-591?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/components/modals/ModalForm.tsx
+++ b/src/components/modals/ModalForm.tsx
@@ -24,14 +24,18 @@ import { TextNodeOptional } from '~types/utils';
 import { Button, ButtonVariety } from '../buttons';
 import { Form, FormHeaderProps, FormRootProps } from '../form';
 import { Modal, ModalProps } from './Modal';
+import { ModalAlert, ModalAlertProps } from './ModalAlert';
+
+type ExcludedModalProps =
+  | 'content'
+  | 'isOpen'
+  | 'onOpenChange'
+  | 'primaryButton'
+  | 'secondaryButton';
 
 type FormRootPropsSubset = Omit<FormRootProps, 'children' | 'onSubmit'>;
-type ModalPropsSubset = Omit<
-  ModalProps,
-  'primaryButton' | 'secondaryButton' | 'content' | 'isOpen' | 'onOpenChange'
->;
 
-export interface ModalFormProps extends FormRootPropsSubset, ModalPropsSubset {
+interface ModalFormBaseProps extends FormRootPropsSubset {
   /**
    * The content of the `Form`, can contain multiple `Form.Section` with their fields
    */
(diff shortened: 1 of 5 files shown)


src/components/modals/ModalForm.tsx
```diff
@@ -137,8 +152,12 @@ export const ModalForm = forwardRef<HTMLDivElement, ModalFormProps>((props, ref)
   const defaultId = `${useId()}modal-form`;
   const formId = id ?? defaultId;
 
+  const ModalComponent = isDestructive ? ModalAlert : Modal;
+
   return (
-    <Modal
+    <ModalComponent
+      // description prop is mandatory for ModalAlert but not for Modal, force typing to the most discriminant type to avoid duplicating a lot of code
+      description={description as ModalAlertPropsSubset['description']}
```
gregaubert, line 160: There is a way to make this typing work correctly by doing something like:

```ts
if (modalProps.isDestructive) {
  const { isDestructive, ...modalAlertProps} = modalProps;
  return <ModalAlert {...modalAlertProps} .../>
}

return <Modal {...modalProps} .../>
```

Doing this correctly switches the type of `modalProps` based on the value of `isDestructive`. But then it implies duplicating all the code that we pass to the Modal components... So I figured I prefer not have the duplication and force a type. But I'm open to suggestions if you have better ideas!

daniel-nagy (approved):
LGTM

src/components/modals/ModalForm.tsx
jeremy-davis-sonarsource: Maybe a stupid question, but can't we use a ModalAlert all the time?
Semantically it makes less sense, but if it provides the structure we need :shrug:

jeremy-davis-sonarsource (approved):

src/components/modals/ModalForm.tsx
gregaubert: I don't think we should, because the modal alert also comes with a different role and interrupts the user.

PR #273, opened 2025-02-12
ECHOES-551 Navbar container

robert-hoover-sonarsource:
[ECHOES-551](https://sonarsource.atlassian.net/browse/ECHOES-551)

ECHOES-551
Part of ECHOES-550

[ECHOES-551]: https://sonarsource.atlassian.net/browse/ECHOES-551?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/components/navbar/Navbar.tsx
+++ b/src/components/navbar/Navbar.tsx
@@ -20,12 +20,25 @@
 
 import { forwardRef } from 'react';
 import styled from '@emotion/styled';
+import { useIntl } from 'react-intl';
 
-const NavbarRoot = forwardRef<HTMLElement, React.PropsWithChildren>((props, ref) => {
-  const { children } = props;
+export interface NavbarProps extends React.PropsWithChildren {
+  className?: string;
+  ariaLabel?: string;
+}
+
+export const NavbarRoot = forwardRef<HTMLElement, NavbarProps>((props, ref) => {
+  const { children, ariaLabel, ...rest } = props;
+  const intl = useIntl();
+
+  const defaultAriaLabel = intl.formatMessage({
+    id: 'navbar.main.navigation',
+    defaultMessage: 'Main navigation',
+    description: 'ARIA-label for the main navigation',
+  });
 
   return (
-    <NavbarContainer aria-label="main navigation" ref={ref} {...props}>
+    <NavbarContainer aria-label={ariaLabel ?? defaultAriaLabel} ref={ref} {...rest}>
       {children}
     </NavbarContainer>
   );
@@ -37,31 +50,25 @@ const NavbarContainer = styled.nav`
   height: 56px;
   display: flex;
   justify-content: space-between;
-  min-width: 1280px; // needs to be replaced with a token?? will sync with Marcio
   align-items: center;
 
-  padding: 0 var(--echoes-dimension-space-300) 0 var(--echoes-dimension-space-0);
+  padding-right: var(--echoes-dimension-space-300);
 
   background-color: var(--echoes-color-background-default);
   border-bottom: var(--echoes-border-width-default) solid var(--echoes-color-border-weak);
 `;
 
-const NavbarLeft = styled.div`
+export const NavbarPrimary = styled.div`
   display: flex;
   align-items: center;
   height: 100%;
 
   gap: var(--echoes-dimension-space-100);
 `;
-const NavbarRight = styled.div`
+export const NavbarSecondary = styled.div`
   display: flex;
   align-items: center;
   height: 100%;
 
   gap: var(--echoes-dimension-space-100);
 `;
-
-export const Navbar = Object.assign(NavbarRoot, {
-  Left: NavbarLeft,
-  Right: NavbarRight,
-});
--- a/src/components/navbar/Navbar.tsx
+++ b/src/components/navbar/Navbar.tsx
@@ -1,0 +1,62 @@
+/*
+ * Echoes React
+ * Copyright (C) 2023-2025 SonarSource SA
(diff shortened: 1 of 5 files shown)


src/components/navbar/Navbar.tsx
```diff
@@ -0,0 +1,67 @@
...
+  );
+});
+
+NavbarRoot.displayName = 'Navbar';
+
+const NavbarContainer = styled.nav`
+  height: 56px;
+  display: flex;
+  justify-content: space-between;
+  min-width: 1280px; // needs to be replaced with a token?? will sync with Marcio
```
jeremy-davis-sonarsource, line 40: Discussed with Marcio: we can remove this and leave it to the products to define the size.
We should only make sure it doesn't collapse onto itself (has an auto min-width based on its contents), but I think flex handles it fine

src/components/navbar/Navbar.tsx
```diff
@@ -0,0 +1,67 @@
...
+ */
+
+import { forwardRef } from 'react';
+import styled from '@emotion/styled';
+
+const NavbarRoot = forwardRef<HTMLElement, React.PropsWithChildren>((props, ref) => {
+  const { children } = props;
+
+  return (
+    <NavbarContainer aria-label="main navigation" ref={ref} {...props}>
```
jeremy-davis-sonarsource, line 28: aria-label needs a default value (translated!) and be overridable (optional prop).

src/components/navbar/Navbar.tsx
```diff
@@ -0,0 +1,67 @@
...
+ *
+ * You should have received a copy of the GNU Lesser General Public License
+ * along with this program; if not, write to the Free Software Foundation,
+ * Inc., 51 Franklin Street, Fifth Floor, Boston, MA  02110-1301, USA.
+ */
+
+import { forwardRef } from 'react';
+import styled from '@emotion/styled';
+
+const NavbarRoot = forwardRef<HTMLElement, React.PropsWithChildren>((props, ref) => {
```
jeremy-davis-sonarsource, line 24: We need `className` as a prop

src/components/navbar/Navbar.tsx
```diff
@@ -0,0 +1,67 @@
...
+
+  gap: var(--echoes-dimension-space-100);
+`;
+const NavbarRight = styled.div`
+  display: flex;
+  align-items: center;
+  height: 100%;
+
+  gap: var(--echoes-dimension-space-100);
+`;
```
jeremy-davis-sonarsource, line 62: These are called PrimarySide & SecondarySide in Figma, we should align the naming

src/components/navbar/Navbar.tsx
```diff
@@ -0,0 +1,67 @@
...
+NavbarRoot.displayName = 'Navbar';
+
+const NavbarContainer = styled.nav`
+  height: 56px;
+  display: flex;
+  justify-content: space-between;
+  min-width: 1280px; // needs to be replaced with a token?? will sync with Marcio
+  align-items: center;
+
+  padding: 0 var(--echoes-dimension-space-300) 0 var(--echoes-dimension-space-0);
```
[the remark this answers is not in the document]
robert-hoover-sonarsource: Good point, thanks Jay!

src/components/navbar/Navbar.tsx
```diff
@@ -0,0 +1,67 @@
...
+  align-items: center;
+  height: 100%;
+
+  gap: var(--echoes-dimension-space-100);
+`;
+
+export const Navbar = Object.assign(NavbarRoot, {
+  Left: NavbarLeft,
+  Right: NavbarRight,
+});
```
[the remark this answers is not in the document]
robert-hoover-sonarsource: The form example does seem like a better approach. I will update it.

src/components/navbar/Navbar.tsx
robert-hoover-sonarsource: Done.

src/components/navbar/Navbar.tsx
robert-hoover-sonarsource: I have removed the min-width, thanks Jay!

src/components/navbar/Navbar.tsx
robert-hoover-sonarsource: Per my conversation with Marcio, I have changed the name of these two components to be `Navbar.Primary` and `Navbar.Secondary` to be aligned on the naming

src/components/navbar/__tests__/Navbar-test.tsx
```diff
@@ -0,0 +1,56 @@
...
+import { IntlProvider } from 'react-intl';
+
+describe('Navbar', () => {
+  it('should render children inside the Navbar', () => {
+    render(
+      <IntlProvider defaultLocale="en-us" locale="en-us">
+        <Navbar>
+          <div>Test</div>
+        </Navbar>
+      </IntlProvider>,
```
jeremy-davis-sonarsource, line 32: NP: You can use the `render` function from src/common/helpers/test-utils.tsx instead of wrapping it in the IntlProvider yourself

jeremy-davis-sonarsource (approved):
LGTM!

src/components/navbar/__tests__/Navbar-test.tsx
robert-hoover-sonarsource: Fixed it to use the util, thanks Jay

PR #274, opened 2025-02-14
ECHOES-550 Implement NavbarItem & NavbarItemsContainer

jeremy-davis-sonarsource:
[ECHOES-550](https://sonarsource.atlassian.net/browse/ECHOES-550)

[ECHOES-550]: https://sonarsource.atlassian.net/browse/ECHOES-550?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


src/components/navbar/NavbarDropdownItem.tsx
```diff
@@ -0,0 +1,64 @@
...
+      <Button
+        className={className}
+        css={navbarItemStyle}
+        ref={ref}
+        suffix={<IconChevronDown />}
+        variety="default-ghost"
+        {...rest}>
+        {children}
+      </Button>
+    </StyledWrapper>
```
daniel-nagy, line 61: [non-blocking] For form components, whenever a component's style changed based on its state, we would add a data attribute to the element. I think the idea was to give the developer a mechanism to select the element based on its state. Here is an example,

https://github.com/SonarSource/echoes-react/blob/d25fa14828094ef9be61500368dc1b58cb87967b/src/components/form/FormField.tsx#L213-L226

I don't know if we to enforce that or not. I have no strong opinions either way. If it were up to me I would just use inline styles 😅.

src/components/navbar/NavbarItemStyles.tsx
```diff
@@ -0,0 +1,49 @@
...
+  text-decoration-line: var(--echoes-text-decoration-none);
+
+  &:hover {
+    background-color: var(--echoes-color-background-default-hover);
+  }
+
+  &:active {
+    background-color: var(--echoes-color-background-default-active);
+  }
+`;
```
daniel-nagy, line 49: [non-blocking] I wonder if it would be better to make this a styled component, e.g.

```tsx
export const NavbarItemBase = styled.div`
  --hover: var(--echoes-color-text-default);
  display: block;
  ...
`;
```

And then use it like this

```tsx
<NavbarItemBase as={Button} {...rest}>
  {children}
</NavbarItemBase>
```

Or is TypeScript not able to infer props in that case?

src/components/navbar/index.ts
```diff
@@ -33,16 +36,47 @@ import { NavbarRoot, NavbarPrimary, NavbarSecondary } from './Navbar';
...
+ *      <Navbar.Item to="/projects">Projects</Navbar.Item>
+ *      <Navbar.DropdownItem
+ *        items={
+ *          <>
+ *            <DropdownMenu.ItemLink to="/link1"><FormattedMessage id='ext1.name'/></DropdownMenu.ItemLink>
+ *            <DropdownMenu.ItemLink to="/link2"><FormattedMessage id='ext2.name'/></DropdownMenu.ItemLink>
+ *          </>
+ *        }
+ *      ><FormattedMessage id='more'/></Navbar.DropdownItem>
+ *    </Navbar.ItemsContainer>
```
daniel-nagy, line 49: [question] Is `FormattedMessage` inside the `</Navbar.DropdownItem>` sensitive to newlines? I find having it on a single line like that harder to read.

src/components/navbar/NavbarItemsContainer.tsx
```diff
@@ -0,0 +1,57 @@
...
+}
+
+export const NavbarItemsContainer = forwardRef<HTMLDivElement, NavbarItemProps>(
+  ({ children, className, ...radixProps }: Readonly<NavbarItemProps>, ref) => {
+    /*
+     * We need to wrap each child in a NavMenu Item to allow Radix to manage it
+     */
+    const wrappedChildren = Children.map(children, (child) => {
+      return <StyledNavigationMenuItem>{child}</StyledNavigationMenuItem>;
+    });
```
daniel-nagy, line 37: [question] Could this wrapping just take place in the NavMenu item components themselves?

src/components/navbar/NavbarItemsContainer.tsx
```diff
@@ -0,0 +1,57 @@
...
+const StyledNavigationMenuRoot = styled(radixNavigationMenu.Root)`
+  display: flex;
+  align-items: center;
+
+  gap: var(--echoes-dimension-space-100);
+`;
+
+const StyledNavigationMenuItem = styled(radixNavigationMenu.Item)`
+  list-style: none;
+`;
```
daniel-nagy, line 57: I think we may need display names for every styled component, something about making generated class names deterministic, not sure though. cc @gregaubert

daniel-nagy (commented):
Looks good! Left some comments but no blockers from me.

I did notice that there are no storybook controls or documentation for the Navbar story. I also noticed there is no `ol` or `ul` around the list items.

image

src/components/navbar/NavbarItemsContainer.tsx
```diff
@@ -0,0 +1,57 @@
...
+import styled from '@emotion/styled';
+import * as radixNavigationMenu from '@radix-ui/react-navigation-menu';
+import { Children, PropsWithChildren, forwardRef } from 'react';
+
+export interface NavbarItemProps
+  extends PropsWithChildren<radixNavigationMenu.NavigationMenuProps> {
+  className?: string;
+}
+
+export const NavbarItemsContainer = forwardRef<HTMLDivElement, NavbarItemProps>(
```
robert-hoover-sonarsource, line 30: [question] I see in Figma it mentions that this component:

`Can grow up to 8. If more options are needed they should be inside of a 'More' dropdown item. `

Currently, there is nothing preventing a user from placing as many items as they want in the `NavbarItemsContainer` component. Does it make sense for us to handle that logic here in this component? What are your thoughts?

src/components/navbar/NavbarDropdownItem.tsx
robert-hoover-sonarsource, line 1: [question] When using keyboard navigation, if I navigate to the dropdown item and press `Enter` to open it, I can navigate through the items using the `Up` and `Down` arrow keys, but pressing `Tab` does not allow me to navigate through them. Is that OK?

src/components/navbar/NavbarDropdownItem.tsx
robert-hoover-sonarsource: Somewhat related to Dan's comment. From some cursory testing in Storybook, I noticed when the dropdown item is focused, the border bottom color is never updated to `var(--echoes-color-border-accent)`

src/components/dropdown-menu/DropdownMenu.tsx
```diff
@@ -43,7 +43,7 @@ export enum DropdownMenuAlign {
   Start = 'start',
 }
 
-interface DropdownMenuRootProps extends radixDropdownMenu.DropdownMenuTriggerProps {
+export interface DropdownMenuRootProps extends radixDropdownMenu.DropdownMenuTriggerProps {
```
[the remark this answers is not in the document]
jeremy-davis-sonarsource: it's not made public, just exported for NavbarDropdownItem to extend.
Regardless, I think it makes sense to rename it :+1:

src/components/navbar/NavbarDropdownItem.tsx
jeremy-davis-sonarsource: That makes a lot of sense, I like it!

Robert, I wanted to discuss it, but forgot to raise the point! I'm not 100% on the solution I went with, but we can revisit it later.

src/components/navbar/NavbarItemStyles.tsx
jeremy-davis-sonarsource: The ref would become an issue, I believe

src/components/navbar/NavbarItemsContainer.tsx
jeremy-davis-sonarsource: The issue is with props forwarding: we want to forward all props from both radix and react-router Links, and spread them to two different nodes.
I couldn't find a simple way to handle that from within the *Item components

src/components/navbar/NavbarItemsContainer.tsx
jeremy-davis-sonarsource: Yeah, this is a UX Designer guideline

src/components/navbar/NavbarDropdownItem.tsx
jeremy-davis-sonarsource: Yes, that is the recommended behavior in the WCAG

jeremy-davis-sonarsource:
> I did notice that there are no storybook controls or documentation for the Navbar story. I also noticed there is no `ol` or `ul` around the list items.

Nice catch! I consider this a blocker. Then again, I'm happy you didn't "request changes" since you're off today :sweat_smile:

PR #279, opened 2025-02-19
ECHOES-550 Fix issues with GlobalNavigation

jeremy-davis-sonarsource:
- Removes the children wrapping logic in the Item Container and wraps the Item and DownloadItem directly
  - This allows fragments around Items in the products
  - Code is simpler and not as brittle. Also no more "magic"
- Add `reloadDocument` prop to Home

Based on the WIP [integration in SQC](https://github.com/SonarSource/sonarcloud-webapp/tree/jay/navbar-migration-sqc)

### DIFF
--- a/src/components/global-navigation/GlobalNavigationDropdownItem.tsx
+++ b/src/components/global-navigation/GlobalNavigationDropdownItem.tsx
@@ -18,6 +18,7 @@
  * Inc., 51 Franklin Street, Fifth Floor, Boston, MA  02110-1301, USA.
  */
 
+import * as radixNavigationMenu from '@radix-ui/react-navigation-menu';
 import { Children, forwardRef, isValidElement, ReactNode, useMemo } from 'react';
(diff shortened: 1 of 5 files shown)


src/components/global-navigation/GlobalNavigationDropdownItem.tsx
```diff
@@ -17,7 +17,7 @@
  * along with this program; if not, write to the Free Software Foundation,
  * Inc., 51 Franklin Street, Fifth Floor, Boston, MA  02110-1301, USA.
  */
-
+import * as radixNavigationMenu from '@radix-ui/react-navigation-menu';
```
david-cho-lerat-sonarsource, line 20:
```suggestion

import * as radixNavigationMenu from '@radix-ui/react-navigation-menu';
```

robert-hoover-sonarsource (approved):
LGTM

PR #288, opened 2025-02-24
ECHOES-593 & ECHOES-594 Add breadcrumb components

david-cho-lerat-sonarsource:
[ECHOES-593](https://sonarsource.atlassian.net/browse/ECHOES-593)
[ECHOES-594](https://sonarsource.atlassian.net/browse/ECHOES-594)

[ECHOES-593]: https://sonarsource.atlassian.net/browse/ECHOES-593?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ
[ECHOES-594]: https://sonarsource.atlassian.net/browse/ECHOES-594?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/components/breadcrumbs/BreadcrumbLink.tsx
+++ b/src/components/breadcrumbs/BreadcrumbLink.tsx
@@ -1,0 +1,59 @@
+/*
+ * Echoes React
+ * Copyright (C) 2023-2025 SonarSource SA
+ * mailto:info AT sonarsource DOT com
+ *
+ * This program is free software; you can redistribute it and/or
+ * modify it under the terms of the GNU Lesser General Public
+ * License as published by the Free Software Foundation; either
+ * version 3 of the License, or (at your option) any later version.
+ *
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the GNU
+ * Lesser General Public License for more details.
+ *
+ * You should have received a copy of the GNU Lesser General Public License
+ * along with this program; if not, write to the Free Software Foundation,
+ * Inc., 51 Franklin Street, Fifth Floor, Boston, MA  02110-1301, USA.
+ */
+
+import styled from '@emotion/styled';
+import { forwardRef } from 'react';
+import { truncate } from '~common/helpers/styles';
+import { LinkHighlight } from '../links';
+import { LinkStandalone, LinkStandaloneProps } from '../links/LinkStandalone';
+
+export interface BreadcrumbLinkProps extends Omit<LinkStandaloneProps, 'children'> {
+  hasEllipsis?: boolean;
+  linkElement: React.ReactNode;
+}
+
+const BreadcrumbLinkContainer = styled.span`
+  max-width: var(--echoes-sizes-breadcrumbs-max-width-default);
+
+  ${truncate}
+`;
+
+const BreadcrumbLinkBase = forwardRef<HTMLAnchorElement, BreadcrumbLinkProps>((props, ref) => {
+  const { hasEllipsis, linkElement, ...linkProps } = props;
+
+  const title = hasEllipsis ? (linkElement as string) : undefined;
+
+  const Link = (
+    <LinkStandalone {...linkProps} highlight={LinkHighlight.Subdued} ref={ref} title={title}>
+      {linkElement}
+    </LinkStandalone>
+  );
+
+  return hasEllipsis ? <BreadcrumbLinkContainer>{Link}</BreadcrumbLinkContainer> : Link;
+});
+
+BreadcrumbLinkBase.displayName = 'BreadcrumbLinkBase';
+
+export const BreadcrumbLink = styled(BreadcrumbLinkBase)`
+  font: var(--echoes-typography-text-small-regular);
+  white-space: nowrap;
+`;
+
+BreadcrumbLink.displayName = 'BreadcrumbLink';
--- a/src/components/breadcrumbs/Breadcrumbs.tsx
+++ b/src/components/breadcrumbs/Breadcrumbs.tsx
@@ -1,0 +1,61 @@
+/*
+ * Echoes React
+ * Copyright (C) 2023-2025 SonarSource SA
+ * mailto:info AT sonarsource DOT com
+ *
+ * This program is free software; you can redistribute it and/or
+ * modify it under the terms of the GNU Lesser General Public
+ * License as published by the Free Software Foundation; either
+ * version 3 of the License, or (at your option) any later version.
+ *
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the GNU
+ * Lesser General Public License for more details.
+ *
+ * You should have received a copy of the GNU Lesser General Public License
+ * along with this program; if not, write to the Free Software Foundation,
+ * Inc., 51 Franklin Street, Fifth Floor, Boston, MA  02110-1301, USA.
+ */
+
+import styled from '@emotion/styled';
+import React, { forwardRef } from 'react';
+import { IconSlash } from '../icons';
+import { BreadcrumbLink, BreadcrumbLinkProps } from './BreadcrumbLink';
+
+export interface BreadcrumbsProps {
+  className?: string;
+  items: BreadcrumbLinkProps[];
+}
+
+const BreadcrumbsBase = forwardRef<HTMLDivElement, BreadcrumbsProps>((props, ref) => {
+  const { items, ...rest } = props;
+
+  return (
+    <div ref={ref} {...rest}>
+      {items.map((item, index) => (
+        <React.Fragment key={`${index}-${item.to.toString()}`}>
+          {index > 0 && <IconSlash color="echoes-color-icon-subdued" />}
+
+          {index === items.length - 1 ? (
+            <span className={item.className}>{item.linkElement}</span>
+          ) : (
+            <BreadcrumbLink {...item} />
+          )}
+        </React.Fragment>
+      ))}
+    </div>
+  );
+});
+
+BreadcrumbsBase.displayName = 'BreadcrumbsBase';
+
+export const Breadcrumbs = styled(BreadcrumbsBase)`
+  display: flex;
+  flex-wrap: wrap;
+  font: var(--echoes-typography-text-small-regular);
+  gap: var(--echoes-dimension-space-75);
+  max-width: 100%;
+`;
+
+Breadcrumbs.displayName = 'Breadcrumbs';
--- a/src/components/breadcrumbs/index.ts
+++ b/src/components/breadcrumbs/index.ts
@@ -18,25 +18,7 @@
  * Inc., 51 Franklin Street, Fifth Floor, Boston, MA  02110-1301, USA.
  */
 
-export * from './breadcrumbs';
-export * from './buttons';
-export * from './checkbox';
-export * from './checkbox-group';
-export * from './dropdown-menu';
-export * from './echoes-provider';
-export * from './form';
-export * from './global-navigation';
-export * from './icons';
-export * from './links';
-export * from './logos';
-export * from './messages';
-export * from './modals';
-export * from './popover';
-export * from './radio-button-group';
-export * from './select';
-export * from './spinner';
-export * from './text-area';
-export * from './text-input';
-export * from './toggle-tip';
-export * from './tooltip';
-export * from './typography';
+export { Breadcrumbs } from './Breadcrumbs';
+
+export type { BreadcrumbLinkProps } from './BreadcrumbLink';
+export type { BreadcrumbsProps } from './Breadcrumbs';
--- a/stories/Breadcrumbs-stories.tsx
+++ b/stories/Breadcrumbs-stories.tsx
@@ -1,0 +1,114 @@
+/*
+ * Echoes React
+ * Copyright (C) 2023-2025 SonarSource SA
+ * mailto:info AT sonarsource DOT com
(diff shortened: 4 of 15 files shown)


src/components/breadcrumbs/BreadcrumbEllipsis.tsx
```diff
@@ -0,0 +1,68 @@
...
+  };
+
+export interface BreadcrumbEllipsisProps extends Pick<DropdownMenuProps, 'id'> {
+  ariaLabel?: string;
+  className?: string;
+  items: BreadcrumbEllipsisItemProps[];
+}
+
+const itemRenderer = ({ iconLeft, linkElement, to, ...rest }: BreadcrumbEllipsisItemProps) => (
+  <DropdownMenu.ItemLink key={crypto.randomUUID()} prefix={iconLeft} to={to} {...rest}>
```
gregaubert, line 42: Let's use an array index intead of a randomUUID? Or let's force to pass something that can serve as an id.

src/components/breadcrumbs/BreadcrumbLink.tsx
```diff
@@ -0,0 +1,82 @@
...
+
+  return hasEllipsis ? <BreadcrumbLinkContainer>{Link}</BreadcrumbLinkContainer> : Link;
+});
+
+BreadcrumbLinkBase.displayName = 'BreadcrumbLinkBase';
+
+export const BreadcrumbLink = styled(BreadcrumbLinkBase)`
+  font: var(--echoes-typography-text-default-regular);
+  font-size: var(--echoes-font-size-10);
+  line-height: var(--echoes-line-height-10);
```
gregaubert, line 78: Can't you use `echoes-typography-text-small-regular` which already contains size-10 and line-height-10 instead of using another typography and changing it's size?

Also if this is set on the BreadcrumbList then it should be inherited by all links inside.

src/components/breadcrumbs/BreadcrumbLink.tsx
```diff
@@ -0,0 +1,82 @@
...
+export type BreadcrumbLinkProps = (
+  | (LinkProps & { isChild?: never })
+  | (Omit<LinkProps, 'to'> & { isChild: true; to?: '' })
+) & {
+  hasEllipsis?: boolean;
+};
+
+const BreadcrumbLinkContainer = styled.span`
+  font-size: var(--echoes-font-size-10);
+  line-height: var(--echoes-line-height-10);
```
gregaubert, line 40: Use `echoes-typography-text-small-regular` instead. And you probably don't even need it, since you already define it on the BreadcrumbLink

src/components/links/LinkBase.tsx
```diff
@@ -81,6 +91,14 @@ export const LinkBase = forwardRef<HTMLAnchorElement, LinkProps>((props, ref) =>
     [onClick, shouldBlurAfterClick, shouldPreventDefault, shouldStopPropagation],
   );
 
+  if (isDisabled) {
+    return (
+      <span {...restAndRadixProps} ref={ref} style={style}>
+        {children}
+      </span>
+    );
+  }
```
gregaubert, line 100: I'm still very skeptical that we should provide the `isDisabled` prop on our Link component... I'm not sure what you are gaining from doing this? The style? But what style? Font-size?

I think you should just use a Text component in BreadCrumbLink instead.

src/components/breadcrumbs/BreadcrumbList.tsx
```diff
@@ -0,0 +1,78 @@
...
+        } else {
+          breadcrumb = (
+            <BreadcrumbLink
+              {...(index === items.length - 1 ? { ...item, isChild: true, to: '' } : { ...item })}
+            />
+          );
+        }
+
+        return (
+          <React.Fragment key={crypto.randomUUID()}>
```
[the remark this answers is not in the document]
gregaubert: What copilot said, let's use the array index or let's force devs to provide a stable key.

src/components/breadcrumbs/BreadcrumbList.tsx
```diff
@@ -0,0 +1,78 @@
...
+            {Boolean(index) && <IconSlash color="echoes-color-icon-subdued" />}
```
gregaubert, line 57: Let's make it a `index > 0` so that's more obvious?

src/components/breadcrumbs/BreadcrumbList.tsx
```diff
@@ -0,0 +1,78 @@
...
+    </div>
+  );
+});
+
+BreadcrumbListBase.displayName = 'BreadcrumbListBase';
+
+export const BreadcrumbList = styled(BreadcrumbListBase)`
+  display: flex;
+  flex-wrap: wrap;
+  font-size: var(--echoes-font-size-10);
```
gregaubert, line 72: Let's use a full typography design token instead of `font-size` and `line-height` separatly.

src/components/breadcrumbs/BreadcrumbList.tsx
```diff
@@ -0,0 +1,78 @@
...
+import styled from '@emotion/styled';
+import React, { forwardRef } from 'react';
+import { isDefined } from '~common/helpers/types';
+import { IconSlash } from '../icons';
+import { BreadcrumbEllipsis, BreadcrumbEllipsisProps } from './BreadcrumbEllipsis';
+import { BreadcrumbLink, BreadcrumbLinkProps } from './BreadcrumbLink';
+
+export interface BreadcrumbListProps {
+  className?: string;
+  items: (BreadcrumbLinkProps | BreadcrumbEllipsisProps)[];
```
gregaubert, line 30: I think `isChild` from `BreadcrumbLinkProps` should not be exposed to the devs. Unless we really expect them to override this value for an item that's in the middle of the breadcrumb... But I don't think it really make sense...

src/components/breadcrumbs/BreadcrumbList.tsx
```diff
@@ -0,0 +1,78 @@
...
+      {items.map((item, index) => {
+        let breadcrumb;
+
+        if (isEllipsisItem(item)) {
+          breadcrumb = <BreadcrumbEllipsis {...item} />;
+        } else {
+          breadcrumb = (
+            <BreadcrumbLink
+              {...(index === items.length - 1 ? { ...item, isChild: true, to: '' } : { ...item })}
+            />
```
gregaubert, line 51: If you get rid of the `isDisabled` on Link you can probably also simplify this to:

```suggestion
            <BreadcrumbLink
              {...rest}
              isChild={index === items.length - 1}
            />
```

src/components/breadcrumbs/index.ts
```diff
@@ -0,0 +1,27 @@
...
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the GNU
+ * Lesser General Public License for more details.
+ *
+ * You should have received a copy of the GNU Lesser General Public License
+ * along with this program; if not, write to the Free Software Foundation,
+ * Inc., 51 Franklin Street, Fifth Floor, Boston, MA  02110-1301, USA.
+ */
+
+export { BreadcrumbEllipsis } from './BreadcrumbEllipsis';
+export { BreadcrumbLink } from './BreadcrumbLink';
```
gregaubert, line 22: Should these be available to be used directly by devs ? Only `BreadcrumbList` should be right?

stories/breadcrumbs/BreadcrumbLink-stories.tsx
```diff
@@ -0,0 +1,64 @@
...
+ */
+
+/* eslint-disable no-console */
+import type { Meta, StoryObj } from '@storybook/react';
+import { BreadcrumbLink, IconBell } from '../../src/components';
+import { basicWrapperDecorator } from '../helpers/BasicWrapper';
+
+const meta: Meta<typeof BreadcrumbLink> = {
+  component: BreadcrumbLink,
+  title: 'Echoes/Breadcrumbs/BreadcrumbLink',
```
gregaubert, line 28: Let's move this story to the internal ones if we end up not exposing BreadcrumbLink to devs.

gregaubert (changes requested):

src/components/breadcrumbs/BreadcrumbEllipsis.tsx
david-cho-lerat-sonarsource: using an array index triggers a warning as well :/

src/components/breadcrumbs/BreadcrumbList.tsx
david-cho-lerat-sonarsource: but if `isChild` is not exposed they will be forced to always provide a `to` for the last link, which won't be used, so it will be like `to="#"' for instance - not sure if that's better than stating `isChild` explicitely?

src/components/breadcrumbs/index.ts
david-cho-lerat-sonarsource: depends if they want to build their own logic like the existing `<Breadcrumbs>` component in the projects?

src/components/breadcrumbs/BreadcrumbEllipsis.tsx
jeremy-davis-sonarsource: yeah, I'd use array index + `to.toString()`

david-cho-lerat-sonarsource: dropped this component for now

src/components/links/LinkBase.tsx
david-cho-lerat-sonarsource: the whole changes around isDisabled have been dropped

src/components/breadcrumbs/BreadcrumbList.tsx
david-cho-lerat-sonarsource: discussed with @jeremy-davis-sonarsource as well, we won't expose isChild

david-cho-lerat-sonarsource: actually, the logic to make the last element a non-link has now moved to the breadcrumb list, so isChild/isDisabled no longer exist

src/components/breadcrumbs/BreadcrumbList.tsx
david-cho-lerat-sonarsource: isDisabled & isChild both dropped, code simplified

src/components/breadcrumbs/index.ts
david-cho-lerat-sonarsource: removed these exports, we're only exposing the list component

stories/Breadcrumbs-stories.tsx
```diff
@@ -0,0 +1,114 @@
...
+
+/* eslint-disable no-console */
+import styled from '@emotion/styled';
+import type { Meta, StoryObj } from '@storybook/react';
+import { Breadcrumbs, IconStar, IconTriangleUp } from '../src/components';
+import { basicWrapperDecorator } from './helpers/BasicWrapper';
+
+const meta: Meta<typeof Breadcrumbs> = {
+  component: Breadcrumbs,
+  title: 'Echoes/Breadcrumbs/Breadcrumbs',
```
jeremy-davis-sonarsource, line 29: You can remove a level now

stories/Breadcrumbs-stories.tsx
```diff
@@ -0,0 +1,114 @@
...
+
+const defaultCrumbs = [
+  { iconLeft: <IconTriangleUp />, linkElement: 'Top-level link', to: 'be continued' },
+  { linkElement: 'Second-level link', to: 'be or not to be' },
+  {
+    iconLeft: <IconStar color="echoes-color-icon-danger" isFilled />,
+    linkElement: 'My favorite',
+    to: '',
+  },
+];
```
jeremy-davis-sonarsource, line 55: NP: You're using this only in one story, I'd move it directly in it

src/components/breadcrumbs/Breadcrumbs.tsx
```diff
@@ -0,0 +1,66 @@
...
+      ))}
+    </div>
+  );
+});
+
+BreadcrumbsBase.displayName = 'BreadcrumbsBase';
+
+const ChildBreadcrumb = styled(Text)`
+  font: var(--echoes-typography-text-small-regular);
+`;
```
jeremy-davis-sonarsource, line 56: You can replace the ChildBreadcrumb component with a simple span. It will inherit the font from the Breadcrumbs

jeremy-davis-sonarsource (changes requested):
Yay looks good! One simplification left and a couple of NPs for the story

stories/Breadcrumbs-stories.tsx
david-cho-lerat-sonarsource: ah, yes, a leftover from when we had the ellipsis component

src/components/breadcrumbs/index.ts
```diff
@@ -0,0 +1,23 @@
...
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the GNU
+ * Lesser General Public License for more details.
+ *
+ * You should have received a copy of the GNU Lesser General Public License
+ * along with this program; if not, write to the Free Software Foundation,
+ * Inc., 51 Franklin Street, Fifth Floor, Boston, MA  02110-1301, USA.
+ */
+
+export { Breadcrumbs } from './Breadcrumbs';
```
gregaubert, line 21: Let's also export `BreadcrumbsProps`

gregaubert (approved):
Looks good! 👍🏻

src/components/breadcrumbs/index.ts
david-cho-lerat-sonarsource: why? when would you need it?

PR #293, opened 2025-03-10
ECHOES-640 Prevent default behaviour on form submit if no action submited

kevinmlsilva:
[ECHOES-640](https://sonarsource.atlassian.net/browse/ECHOES-640)

Part of 

[ECHOES-640]: https://sonarsource.atlassian.net/browse/ECHOES-640?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/components/form/FormRoot.tsx
+++ b/src/components/form/FormRoot.tsx
@@ -18,7 +18,8 @@
  * Inc., 51 Franklin Street, Fifth Floor, Boston, MA  02110-1301, USA.
  */
 import styled from '@emotion/styled';
-import { FormHTMLAttributes, forwardRef, ReactNode } from 'react';
+import { FormEvent, FormHTMLAttributes, forwardRef, ReactNode, useCallback } from 'react';
+import { isDefined } from '~common/helpers/types';
 
 type FormAttributesSubset =
   | 'action'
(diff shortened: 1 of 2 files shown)


src/components/form/FormRoot.tsx
```diff
@@ -18,7 +18,8 @@
  * Inc., 51 Franklin Street, Fifth Floor, Boston, MA  02110-1301, USA.
  */
 import styled from '@emotion/styled';
-import { FormHTMLAttributes, forwardRef, ReactNode } from 'react';
+import { FormEvent, FormHTMLAttributes, forwardRef, ReactNode, useCallback } from 'react';
+import { isDefined } from '../../common/helpers/types';
```
gregaubert, line 22:
```suggestion
import { isDefined } from '~common/helpers/types';
```

src/components/form/__tests__/Form-test.tsx
```diff
@@ -50,6 +50,14 @@ it('should submit and reset the form', async () => {
   expect(onReset).toHaveBeenCalled();
 });
 
+it('should not prevent default on submit if action is defined', async () => {
```
gregaubert, line 53: The test is raising errors:
image

Let's either expect the error, but seems a bit weird since it's most probably a jest related error more than a real one...

src/components/form/__tests__/Form-test.tsx
gregaubert, line 1: Let's also update the `should submit and reset the form` test to remove the `e.preventDefault()`. Since it should be done automatically it shouldn't throw anymore.

gregaubert (approved):
Looks good!
I just added a commit with a few improvements to the tests.
To ensure we are testing the right thing in the first and that we didn't call preventDefault automatically in the second one.

PR #300, opened 2025-03-21
ECHOES-659 Add new width token (56px) width.700

david-cho-lerat-sonarsource:


jeremy-davis-sonarsource (approved):

jeremy-davis-sonarsource:
I think the title broke the jira automation :sweat_smile:

PR #301, opened 2025-03-21
ECHOES-647,648,649 Implement RatingBadge components

david-cho-lerat-sonarsource:
[ECHOES-647](https://sonarsource.atlassian.net/browse/ECHOES-647)

[ECHOES-647]: https://sonarsource.atlassian.net/browse/ECHOES-647?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/components/badges/RatingBadge.tsx
+++ b/src/components/badges/RatingBadge.tsx
@@ -1,0 +1,109 @@
+/*
+ * Echoes React
+ * Copyright (C) 2023-2025 SonarSource SA
+ * mailto:info AT sonarsource DOT com
+ *
+ * This program is free software; you can redistribute it and/or
+ * modify it under the terms of the GNU Lesser General Public
+ * License as published by the Free Software Foundation; either
+ * version 3 of the License, or (at your option) any later version.
+ *
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the GNU
+ * Lesser General Public License for more details.
+ *
+ * You should have received a copy of the GNU Lesser General Public License
+ * along with this program; if not, write to the Free Software Foundation,
+ * Inc., 51 Franklin Street, Fifth Floor, Boston, MA  02110-1301, USA.
+ */
+
+import styled from '@emotion/styled';
+import { forwardRef } from 'react';
+import { isStringDefined } from '~common/helpers/types';
+
+export enum RatingBadgeRating {
+  A = 'A',
+  B = 'B',
+  C = 'C',
+  D = 'D',
+  E = 'E',
+  Null = '—',
+}
+
+export enum RatingBadgeSize {
+  ExtraSmall = 'xs', // 16px
+  Small = 'sm', // 24px (default)
+  Medium = 'md', // 32px
+  Large = 'lg', // 48px
+  ExtraLarge = 'xl', // 56px
+}
+
+export interface RatingBadgeProps {
+  ariaLabel?: string;
+  className?: string;
+  rating?: `${RatingBadgeRating}`;
+  size?: `${RatingBadgeSize}`;
+  style?: React.CSSProperties;
+}
+
+export const RatingBadge = forwardRef<HTMLDivElement, RatingBadgeProps>(
+  ({ ariaLabel, rating: ratingPropValue, size = RatingBadgeSize.Small, ...htmlProps }, ref) => {
+    const rating = isStringDefined(ratingPropValue) ? ratingPropValue : RatingBadgeRating.Null;
+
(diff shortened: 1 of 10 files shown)


src/components/badges/RatingBadge.tsx
```diff
@@ -0,0 +1,107 @@
...
+  Medium = 'md', // 32px
+  Large = 'lg', // 48px
+  ExtraLarge = 'xl', // 56px
+}
+
+export interface RatingBadgeProps {
+  ariaLabel?: string;
+  className?: string;
+  rating?: RatingBadgeRating;
+  size?: RatingBadgeSize;
```
jeremy-davis-sonarsource, line 45: We've been typing these like this to allow passing the value directly:
```rating?: `${RatingBadgeRating}`;```

File: src/components/badges/RatingBadgeButton.tsx [hunk 2]
```diff
@@ -0,0 +1,92 @@
...
+RatingBadgeButton.displayName = 'RatingBadgeLink';
+
+const RatingBadgeButtonStyled = styled(RatingBadgeButtonInner)`
+  border-radius: var(--echoes-border-radius-full);
+  padding: 0;
+
+  ${({ rating }) =>
+    rating === RatingBadgeRating.Null
+      ? ''
+      : `
```
jeremy-davis-sonarsource, line 70: I'm not sure I understand the use for `RatingBadgeButtonInner`. You can `styled(Button)<SpecificProps>`.
This should also solve the issue of the optional size prop, since you can force it at this point

src/components/badges/RatingBadgeButton.tsx
```diff
@@ -0,0 +1,92 @@
...
+
+RatingBadgeButtonInner.displayName = 'RatingBadgeButtonInner';
+
+export const RatingBadgeButton = forwardRef<HTMLButtonElement, RatingBadgeButtonProps>(
+  (
+    { className, rating: ratingPropValue, size = RatingBadgeSize.Medium, style, ...buttonProps },
+    ref,
+  ) => {
+    // below: do not replace || with ?? as we want to handle ""
+    const rating = ratingPropValue || RatingBadgeRating.Null;
```
jeremy-davis-sonarsource, line 48: NP: you could make it explicit witha ternary instead:
`isDefined(ratingPropValue) ? ratingPropValue : RatingBadgeRating.Null`, right?

jeremy-davis-sonarsource (changes requested):
Nice work!
Just have doubts about the button variant

src/components/badges/RatingBadgeButton.tsx
david-cho-lerat-sonarsource: it would actually be `isStringDefined`, but yes, that's a good point - I'll do the change

src/components/badges/RatingBadgeButton.tsx
david-cho-lerat-sonarsource: that's interesting but how do I override Button's size prop? Doing as you suggests still gives me a TS error because I'm passing a RatingBadgeSize where a Button size is expected (size?: `${ButtonSize}`;)

david-cho-lerat-sonarsource: yes, too bad - there may be a better way to do this but I'm not aware of it. This could be improved later.

src/components/badges/RatingBadgeButton.tsx
```diff
@@ -0,0 +1,90 @@
...
+
+const RatingBadgeButtonStyled = styled(RatingBadgeButtonInner)`
+  border-radius: var(--echoes-border-radius-full);
+  padding: 0;
+
+  ${({ rating = RatingBadgeRating.Null }) =>
+    rating === RatingBadgeRating.Null
+      ? ''
+      : `
+          & div: hover {
```
robert-hoover-sonarsource, line 71:
```suggestion
          & div:hover {
```

robert-hoover-sonarsource (approved):
Great work! I left a couple of small suggestions.

src/components/badges/RatingBadgeButton.tsx
```diff
@@ -0,0 +1,90 @@
...
+        ref={ref}
+        variety={ButtonVariety.DefaultGhost}
+        {...{ className, rating, size, style, ...buttonProps }}>
+        <RatingBadge {...{ rating, size }} />
+      </RatingBadgeButtonStyled>
+    );
+  },
+);
+
+RatingBadgeButton.displayName = 'RatingBadgeLink';
```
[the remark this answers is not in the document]
david-cho-lerat-sonarsource: ooh well done, another copy-waste

jeremy-davis-sonarsource (approved):

PR #303, opened 2025-03-26
ECHOES-654 WIP implement card component

robert-hoover-sonarsource:
[ECHOES-654](https://sonarsource.atlassian.net/browse/ECHOES-654)

Part of 

[ECHOES-654]: https://sonarsource.atlassian.net/browse/ECHOES-654?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


src/components/card/CardBody.tsx
```diff
@@ -0,0 +1,57 @@
...
+const paddingMap = {
+  [CardSize.Small]: 'var(--echoes-dimension-space-150)',
+  [CardSize.Medium]: 'var(--echoes-dimension-space-200)',
+  [CardSize.Large]: 'var(--echoes-dimension-space-300)',
+};
+
+export const CardBody = React.forwardRef<HTMLDivElement, Readonly<CardBodyProps>>(
+  ({ children, className, insetContent = false, size = CardSize.Medium }, ref) => {
+    return (
+      <StyledBody className={className} insetContent={insetContent} ref={ref} size={size}>
```
david-cho-lerat-sonarsource, line 41: NP: could be: `{...{ className, insetContent, ref, size}}`

src/components/card/CardBody.tsx
```diff
@@ -0,0 +1,57 @@
...
+
+const paddingMap = {
+  [CardSize.Small]: 'var(--echoes-dimension-space-150)',
+  [CardSize.Medium]: 'var(--echoes-dimension-space-200)',
+  [CardSize.Large]: 'var(--echoes-dimension-space-300)',
+};
+
+export const CardBody = React.forwardRef<HTMLDivElement, Readonly<CardBodyProps>>(
+  ({ children, className, insetContent = false, size = CardSize.Medium }, ref) => {
+    return (
```
david-cho-lerat-sonarsource, line 40: you don't need `=> { return (<StyleBody`, this is an arrow function, just use `=> (<StyleBody`

src/components/card/index.ts
```diff
@@ -0,0 +1,30 @@
...
+ * Inc., 51 Franklin Street, Fifth Floor, Boston, MA  02110-1301, USA.
+ */
+
+import { CardBody } from './CardBody';
+import { CardHeader } from './CardHeader';
+import { CardRoot } from './CardRoot';
+
+export { CardSize } from './CardRoot';
+
+export const Card = Object.assign(CardRoot, {
```
david-cho-lerat-sonarsource, line 27: `Object.assign`??

robert-hoover-sonarsource: This is a pattern that was used in the Form component, and Jay and I also used in the GlobalNavigation. I believe the benefit is to have all of the TSDocs in one location. @jeremy-davis-sonarsource is that right?

david-cho-lerat-sonarsource: OK, but why `Object.assign`, though? Wouldn't `export const Card = { ...CardRoot, Header: CardHeader,` work?

src/components/card/CardTypes.ts
```diff
@@ -0,0 +1,25 @@
...
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the GNU
+ * Lesser General Public License for more details.
+ *
+ * You should have received a copy of the GNU Lesser General Public License
+ * along with this program; if not, write to the Free Software Foundation,
+ * Inc., 51 Franklin Street, Fifth Floor, Boston, MA  02110-1301, USA.
+ */
+
+export enum CardSize {
```
david-cho-lerat-sonarsource, line 21: the enum is called CardSize and the file CardTypes.ts, this is slightly confusing

david-cho-lerat-sonarsource: it's not a type per se, and there's only one export, maybe the file name should be CardSize.ts, or even there shouldn't be a file at all and the enum should be part of the main Card file, like we did for other components?

src/components/card/CardRoot.tsx
```diff
@@ -0,0 +1,47 @@
...
+
+export interface CardProps {
+  children: React.ReactNode;
+  className?: string;
+  size?: CardSize;
+}
+
+const CardContext = createContext<CardSize>(CardSize.Medium);
+
+export function useCardContext() {
```
david-cho-lerat-sonarsource, line 33: there's only one piece of information: the card size. I'd say this should be called useCardSize

src/components/card/CardHeader.tsx
```diff
@@ -0,0 +1,104 @@
...
+            }),
+            [sizeConfig.styles],
+          )}
+          ref={ref}>
+          <CardHeaderContentStyled>
+            <CardHeaderTextStyled>
+              <Heading as={sizeConfig.headingLevel}>{title}</Heading>
+              {rightContent && <RightContentStyled>{rightContent}</RightContentStyled>}
+            </CardHeaderTextStyled>
+            {description && <DescriptionStyled>{description}</DescriptionStyled>}
```
david-cho-lerat-sonarsource, line 74:
```suggestion

            {description && <DescriptionStyled>{description}</DescriptionStyled>}
```

src/components/card/CardHeader.tsx
```diff
@@ -0,0 +1,104 @@
...
+          </CardHeaderContentStyled>
+        </CardHeaderStyled>
+        {hasDivider && <Divider />}
```
david-cho-lerat-sonarsource, line 77:
```suggestion

        {hasDivider && <Divider />}
```

stories/Card-stories.tsx
```diff
@@ -0,0 +1,245 @@
...
+};
+
+export const NoDescription: Story = {
+  args: {
+    size: CardSize.Medium,
+  },
+  render: (args) => (
+    <div style={{ display: 'flex', flexDirection: 'column', gap: '20px', padding: '40px' }}>
+      <Card {...args}>
+        <Card.Header hasDivider title="Header Without Description" />
```
david-cho-lerat-sonarsource, line 195: Shouldn't the header be vertically aligned automatically in this case? The spacing looks weid, here:

image

stories/Card-stories.tsx
```diff
@@ -0,0 +1,245 @@
...
+  },
+  render: (args) => (
+    <div style={{ display: 'flex', flexDirection: 'column', gap: '20px', padding: '40px' }}>
+      <Card {...args}>
+        <Card.Header
+          description="This card has a title and description but no divider between header and body"
+          title="Card Without Divider"
+        />
+        <Card.Body insetContent>
+          <div style={{ padding: '0 16px' }}>
```
david-cho-lerat-sonarsource, line 235: again, we shouldn't need an extra layer here, but apply whatever extra style to the parent element

stories/Card-stories.tsx
```diff
@@ -0,0 +1,245 @@
...
+            <Text as="div">
+              Without a divider, the header and body content flow together visually. This works well
```
david-cho-lerat-sonarsource, line 237: in all the other examples, the card content seem vertically align, but not for this one. Is there a reason for this?:

image

david-cho-lerat-sonarsource (changes requested):
Good job, but I still have a few questions/concerns :sweat_smile:

src/components/card/CardRoot.tsx
robert-hoover-sonarsource: That makes sense. I'll change the naming.

src/components/card/CardTypes.ts
robert-hoover-sonarsource: Yes, I think having it as part of the CardRoot is better. I'll update this.

src/components/card/index.ts
robert-hoover-sonarsource: That also works :)

robert-hoover-sonarsource: Actually, I take that back. When I don't use Object.assign, storybook breaks. My assumption is it's because Object.assign preserves the function identity. So when we create a plain object, although it contains all of CardRoots enumerable properties, it loses its identity as a React component function.

david-cho-lerat-sonarsource: huh, that's a shame, `Object.assign` feels outdated :sweat_smile:

src/components/card/CardTypes.ts
david-cho-lerat-sonarsource: I don't know where the circular dependency is, but having it in CardStyles sounds wrong as it's not a style. Keeping it as a separate file with a relevant name is fine then I guess :+1:

stories/Card-stories.tsx
robert-hoover-sonarsource: No particular reason. I'll change it to be consistent with the other examples.

src/components/card/CardTypes.ts
robert-hoover-sonarsource: Sounds good!

david-cho-lerat-sonarsource (approved):
Looks great, godd job! :muscle:

PR #312, opened 2025-04-23
ECHOES-695 Change border token from severity badges critical to blocker

jeremy-davis-sonarsource:
[ECHOES-695](https://sonarsource.atlassian.net/browse/ECHOES-695)

Part of ECHOES-686

[ECHOES-695]: https://sonarsource.atlassian.net/browse/ECHOES-695?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


david-cho-lerat-sonarsource (approved):
LGTM but I would rename it to be more explicit: we're not just "changing critical to blocker" but doing the opposite as well, so maybe "ECHOES-695 Swap 'blocker' and 'critical' border colors for severity badges" or something?

PR #314, opened 2025-04-24
ECHOES-686 New BadgeSeverity component

jeremy-davis-sonarsource:
[ECHOES-686](https://sonarsource.atlassian.net/browse/ECHOES-686)

[ECHOES-686]: https://sonarsource.atlassian.net/browse/ECHOES-686?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


src/components/badges/BadgeSeverity.tsx
```diff
@@ -0,0 +1,250 @@
...
+  Low = 'low',
+  Info = 'info',
+}
+
+type excludedButtonProps = 'isLoading' | 'size' | 'variety';
+type InheritedButtonProps = Omit<ButtonCommonProps, excludedButtonProps> &
+  Pick<ButtonIconProps, 'isIconFilled' | 'tooltipContent' | 'tooltipOptions'>;
+
+export interface BadgeSeverityProps extends InheritedButtonProps {
+  PrefixIcon?: React.ForwardRefExoticComponent<
```
david-cho-lerat-sonarsource, line 52: I understand this is capitalized because it is a component, but we're inconsistent with, e.g. `iconLeft` in `LinkStandalone`, both with the name and the uppercase - how about calling it `iconLeft` here as well?
Or, yes, `prefixIcon` but all lowercase, which would look like the `prefix` in `DropDownMenuItem` - WDYT?

src/components/badges/BadgeSeverity.tsx
```diff
@@ -0,0 +1,250 @@
...
+  IconSeverityBlocker,
+  IconSeverityCritical,
+  IconSeverityHigh,
+  IconSeverityInfo,
+  IconSeverityLow,
+} from '../icons';
+import { IconFilledProps } from '../icons/IconWrapper';
+import { Tooltip } from '../tooltip';
+
+export enum BadgeSeverityLevel {
```
david-cho-lerat-sonarsource, line 39: do we want to sort these?

src/components/badges/BadgeSeverity.tsx
```diff
@@ -0,0 +1,250 @@
...
+  quality: string;
+  severity: `${BadgeSeverityLevel}`;
+}
+
+export const BadgeSeverity = forwardRef<HTMLButtonElement, BadgeSeverityProps>((props, ref) => {
+  const intl = useIntl();
+
+  const defaultAriaLabel = intl.formatMessage({
+    id: 'badge_severity.click_to_edit',
+    defaultMessage: 'Change the severity',
```
david-cho-lerat-sonarsource, line 65: do we still use `defaultMessage`? We got rid of it in SQC/SQS

src/components/badges/BadgeSeverity.tsx
jeremy-davis-sonarsource: I'm not sure: it's a scale, this is the scale order

src/components/badges/BadgeSeverity.tsx
jeremy-davis-sonarsource: We need it in echoes: that's how we generate the translations

src/components/badges/BadgeSeverity.tsx
david-cho-lerat-sonarsource: but about the capitalization: `iconLeft` in `LinkStandalone` expects a component - you're saying it should then be called `IconLeft`?

david-cho-lerat-sonarsource: sorry, `iconLeft` in `LinkStandalone` expects a rendered element, not a component - thanks for the clarification call!

david-cho-lerat-sonarsource:
There seems to be a contrast issue with the dark theme info icon?  The "i" is not visible enough against its background:

image

image

jeremy-davis-sonarsource:
Discussed with Marcio, it doesn't need to be readable. Also, the contrast tool in storybook seems happy with the contrast?

src/components/badges/__tests__/BadgeSeverity-test.tsx
```diff
@@ -0,0 +1,54 @@
...
+ */
+
+import { screen } from '@testing-library/react';
+import { render } from '~common/helpers/test-utils';
+import { IconBug } from '../../../components/icons';
+import { BadgeSeverity } from '../BadgeSeverity';
+
+describe('BadgeSeverity', () => {
+  it('renders as expected', () => {
+    render(<BadgeSeverity data-testid="badge" quality="Thing" severity="info" />);
```
david-cho-lerat-sonarsource, line 28: NP: I now we can use either, but shouldn't we use the enum value for the severity, instead of a string, to showcase it here?

src/components/badges/__tests__/BadgeSeverity-test.tsx
```diff
@@ -0,0 +1,54 @@
...
+        severity="high"
+      />,
+    );
+
+    expect(screen.getByLabelText('custom label')).toBeInTheDocument();
+    expect(screen.getByText('Quality')).toBeInTheDocument();
+  });
+
+  it("should show a loading state, it doesn't prevent clicking", () => {
+    render(<BadgeSeverity ariaLabel="click me" isLoading quality="Some quality" severity="low" />);
```
david-cho-lerat-sonarsource, line 49: NP: same remark about the enum, assuming we want to encourage its use over the string value?

i18n/keys.json
```diff
@@ -1,4 +1,7 @@
 {
+  "badge_severity.click_to_edit": {
```
david-cho-lerat-sonarsource, line 2: NP: as the message is about "change", I believe it should be "change" in the key as well instead of "edit": "badge_severity.click_to_change" for instance?

david-cho-lerat-sonarsource (approved):
LGTM, well done! :muscle: 
(just a couple of NPs because I can't help it, I am a picker of nits)

i18n/keys.json
jeremy-davis-sonarsource: as discussed, making the aria-label mandatory, so we don't need this translation at all

src/components/badges/__tests__/BadgeSeverity-test.tsx
jeremy-davis-sonarsource: As discussed, preserving the `value` approach

PR #316, opened 2025-04-28
ECHOES-685 Fix BadgeCounter size

jeremy-davis-sonarsource:
It seems the badge doesn't adapt well to all situations. Enforcing the height is necessary:

image


david-cho-lerat-sonarsource (approved):

jeremy-davis-sonarsource:
Argh, I just noticed badges can have a string value, like `23+`...

david-cho-lerat-sonarsource (approved):

PR #321, opened 2025-05-02
ECHOES-670 Fix mend vulnerability on @babel/helpers

gregaubert:
Part of


jeremy-davis-sonarsource (commented):
Where is this all coming from? there's no change to `package.json`?

gregaubert:
> Where is this all coming from? there's no change to `package.json`?

Black magic! 🔮

It comes from deleting all the babel related entries in yarn.lock and regenerating that part with `yarn` command to ensure we don't have some old/compatible transitive dependencies.

gregaubert:
I tried this way because babel/helpers is transitive dependency and not directly referenced in package.json and it worked!

jeremy-davis-sonarsource (approved):

PR #326, opened 2025-05-12
ECHOES-722 Allow ToggleButtonGroup to be disabled

jeremy-davis-sonarsource:
[ECHOES-722](https://sonarsource.atlassian.net/browse/ECHOES-722)

[ECHOES-722]: https://sonarsource.atlassian.net/browse/ECHOES-722?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/components/toggle-button-group/ToggleButtonGroup.tsx
+++ b/src/components/toggle-button-group/ToggleButtonGroup.tsx
@@ -41,6 +41,7 @@ export interface ToggleButtonGroupProps {
 
   className?: string;
 
+  isDisabled?: boolean;
   /**
    * Callback function triggered when the selected value changes.
    * @param value - The newly selected value.
@@ -49,14 +50,13 @@ export interface ToggleButtonGroupProps {
 
   /**
    * The currently selected value in the toggle button group.
-   * If undefined, no option is selected.
    */
-  selected: string | undefined;
+  selected: string;
 }
 
 export const ToggleButtonGroup = forwardRef<HTMLDivElement, ToggleButtonGroupProps>(
   (props, ref) => {
-    const { onChange, options, selected, ...additionalProps } = props;
+    const { isDisabled = false, onChange, options, selected, ...additionalProps } = props;
 
     const handleChange = useCallback(
       (value: string) => {
@@ -70,6 +70,7 @@ export const ToggleButtonGroup = forwardRef<HTMLDivElement, ToggleButtonGroupPro
     return (
       <StyledRoot
         {...additionalProps}
+        disabled={isDisabled}
         onValueChange={handleChange}
         ref={ref}
         type="single"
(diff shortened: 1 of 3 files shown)


src/components/toggle-button-group/ToggleButtonGroup.tsx
```diff
@@ -56,7 +57,13 @@ export interface ToggleButtonGroupProps {
 
 export const ToggleButtonGroup = forwardRef<HTMLDivElement, ToggleButtonGroupProps>(
   (props, ref) => {
-    const { onChange, options, selected, ...additionalProps } = props;
+    const {
+      isDisabled = false,
+      onChange,
+      options,
+      selected = options[0].value,
```
david-cho-lerat-sonarsource, line 64: we should also remove `undefined` for `selected` in the props interface

src/components/toggle-button-group/__tests__/ToggleButtonGroup-test.tsx
```diff
@@ -49,6 +49,21 @@ describe('RadioButtonGroup', () => {
...
+  it('should not react to clicking when disabled', async () => {
+    const onChange = jest.fn();
+    const { user } = renderToggleButtonGroup({ isDisabled: true, onChange });
+
+    await user.click(screen.getByRole('radio', { name: 'b' }));
+
+    expect(onChange).not.toHaveBeenCalled();
+  });
+
+  it('should select first option by default', () => {
```
david-cho-lerat-sonarsource, line 62:
```suggestion
  it('should select the first option by default', () => {
```

stories/ToggleButtonGroup-stories.tsx
```diff
@@ -40,7 +40,7 @@ type Story = StoryObj<typeof ToggleButtonGroup>;
 
 export const Default: Story = {
   args: {
-    selected: '',
+    selected: undefined,
```
david-cho-lerat-sonarsource, line 43: shouldn't it be the 1st option instead?

src/components/toggle-button-group/ToggleButtonGroup.tsx
```diff
@@ -148,14 +153,16 @@ const StyledItem = styled(RadixToggleGroup.Item)`
   align-items: center;
 
   font: var(--echoes-typography-text-default-regular);
+  color: var(--echoes-color-text-default);
```
david-cho-lerat-sonarsource, line 156: shouldn't we also have a `cursor: pointer;` like the other buttons?

david-cho-lerat-sonarsource (changes requested):
Nice, I think we should get rid of `undefined` for `selected`, and also probably have a `cursor:pointer`?

src/components/toggle-button-group/ToggleButtonGroup.tsx
jeremy-davis-sonarsource: Yes, that makes more sense. But I'll remove this then

david-cho-lerat-sonarsource (approved):
Cool :+1:

PR #377, opened 2025-06-02
ECHOES-677 Add Spotlight (WIP)

david-cho-lerat-sonarsource:


src/components/spotlight-tour/SpotlightTourModal.tsx
```diff
@@ -0,0 +1,197 @@
...
+  width: 100%;
+`;
+
+CardFooterContainer.displayName = 'CardFooterContainer';
+
+const ImageContainer = styled.div`
+  align-items: center;
+  align-self: stretch;
+  background:
+    radial-gradient(
```
david-cho-lerat-sonarsource, line 172: @marciopmoreira6 is there a token for this?

src/components/spotlight-tour/SpotlightTourModal.tsx
```diff
@@ -0,0 +1,197 @@
...
+ImageContainer.displayName = 'ImageContainer';
+
+const SpotlightTourModalStyled = styled.div`
+  background: var(--echoes-color-background-accent-default);
+  border-radius: var(--echoes-border-radius-400);
+  box-shadow: var(--echoes-box-shadow-large);
+  display: flex;
+  flex-direction: column;
+  max-height: 600px;
+  min-width: 100px;
```
david-cho-lerat-sonarsource, line 193: @marciopmoreira6 are there tokens for this max-height and min-width?

src/components/spotlight/SpotlightModalForStep.tsx
```diff
@@ -0,0 +1,126 @@
...
+     */
+    step: SpotlightStep;
+  };
+
+export function SpotlightModalForStep(props: Readonly<SpotlightModalForStepProps>) {
+  const {
+    activeTargetClassName = 'active',
+    backProps,
+    closeProps,
+    continuous,
```
gregaubert, line 41: isContinuous?

src/components/spotlight/SpotlightModalForStep.tsx
```diff
@@ -0,0 +1,126 @@
...
+              )}
+            </>
+          )}
+        </ButtonGroup>
+      }
+      cardBodyText={
+        <>
+          {step.content}
+
+          {stepLink && <div className="sw-pt-4">{stepLink}</div>}
```
gregaubert, line 116: We don't have Tailwind in Echoes

src/components/spotlight/SpotlightTypes.tsx
```diff
@@ -0,0 +1,73 @@
...
+ * You should have received a copy of the GNU Lesser General Public License
+ * along with this program; if not, write to the Free Software Foundation,
+ * Inc., 51 Franklin Street, Fifth Floor, Boston, MA  02110-1301, USA.
+ */
+
+import { ReactNode } from 'react';
+import { Props as JoyrideProps, Step as JoyrideStep } from 'react-joyride';
+import { TextNode } from '~types/utils';
+
+type SpotlightModalPlacement = 'left' | 'right' | 'top' | 'bottom' | 'center';
```
gregaubert, line 25: We should make it an enum to keep consistency with the rest of the lib.

src/components/spotlight/SpotlightTypes.tsx
```diff
@@ -0,0 +1,73 @@
...
+  /**
+   * The label for the Next button - defaults to "Next" (optional)
+   */
+  nextLabel?: TextNode;
+
+  /**
+   * The label for the Skip button - defaults to "Skip" (optional)
+   */
+  skipLabel?: TextNode;
+
```
gregaubert, line 56: We might want to make them `TextNodeOptional` if we want to also allow passing `false` or `null` when we don't want to have values. It depends on how these are used in the component.

src/components/spotlight/SpotlightModalForStep.tsx
```diff
@@ -0,0 +1,126 @@
...
+    continuous,
+    index,
+    isLastStep,
+    primaryProps,
+    size,
+    skipProps: { 'aria-label': skipPropsAriaLabel, ...skipProps },
+    step,
+    stepLink,
+    stepXofYLabel,
+    tooltipProps,
```
gregaubert, line 50: Isn't it misleading to have something named `tooltipProps` but that doesn't match our Echoes TooltipProps types?

src/components/spotlight/SpotlightModal.tsx
```diff
@@ -0,0 +1,153 @@
...
+
+  /**
+   * The text to display in the card body
+   */
+  cardBodyText: ReactNode;
+
+  /**
+   * The text to display in the card header
+   */
+  cardHeaderText: ReactNode;
```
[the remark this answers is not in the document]
david-cho-lerat-sonarsource: because the Figma element is "Card" so during the implem I used the same terminology

src/components/spotlight/SpotlightModalForStep.tsx
gregaubert: I don't think so, we are doing this in many places to make sure we keep a consistent API of Echoes. For example if you look at the Select where we use mantine we have plenty of props renamed to suit us.

Doing that also allow us to provide proper documentation for these props in our TSDoc if needed.
It also kind of forces us to be exhaustive with the API we expose and it make sure we don't allow un-expected behaviors that could be introduced without us noticing through new props added with a react-joyride upgrade.

src/components/spotlight/SpotlightModalForStep.tsx
david-cho-lerat-sonarsource: I can't completely rename this one from the start as it is passed by react-joyride itself, so we receive it as "tooltipProps", but I can rename it afterwards: `tooltipProps: spotlightModalOtherProps`

src/components/spotlight/SpotlightModal.tsx
```diff
@@ -0,0 +1,153 @@
...
+
+        <CardFooterContainer>{actions}</CardFooterContainer>
+      </CardContainer>
+    </SpotlightModalStyled>
+  ),
+);
+
+SpotlightModal.displayName = 'SpotlightModal';
+
+const CardContainer = styled.div`
```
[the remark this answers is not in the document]
david-cho-lerat-sonarsource: yeah that came from Figma: while doing the implem it was easier to use the same terminology - image, card, etc to match what styles the Figma elements had. But it doesn't have to be this way.

src/components/spotlight/SpotlightModalForStep.tsx
```diff
@@ -0,0 +1,126 @@
...
+  if (!step.hideFooter && (stepXofYLabel || size > 1)) {
+    helperText =
+      stepXofYLabel?.(index + 1, size) ??
+      intl.formatMessage(
+        {
+          defaultMessage: '{0} of {1}',
+          description: 'The helper text showing the step progress on a Spotlight',
+          id: 'spotlight.step_x_of_y',
+        },
+        { '0': index + 1, '1': size },
```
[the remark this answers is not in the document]
david-cho-lerat-sonarsource: yeah we could do that

src/components/spotlight/SpotlightModal.tsx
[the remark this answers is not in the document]
david-cho-lerat-sonarsource: changed to ModalContainer etc

src/components/spotlight/SpotlightModalForStep.tsx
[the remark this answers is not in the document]
david-cho-lerat-sonarsource: changed

src/components/spotlight/Spotlight.tsx
```diff
@@ -0,0 +1,113 @@
...
+ * Inc., 51 Franklin Street, Fifth Floor, Boston, MA  02110-1301, USA.
+ */
+
+import { PropsWithChildren } from 'react';
+import { useIntl } from 'react-intl';
+import ReactJoyride, { TooltipRenderProps } from 'react-joyride';
+import { SpotlightModalForStep } from './SpotlightModalForStep';
+import { SpotlightProps, SpotlightStep } from './SpotlightTypes';
+
+export function Spotlight(props: Readonly<SpotlightProps>) {
```
gregaubert, line 27: Are we missing the forwardRef that would attach to SpotlightModalStep and then to the SpotlightModal?

src/components/spotlight/SpotlightModal.tsx
```diff
@@ -0,0 +1,153 @@
...
+export interface SpotlightModalProps {
+  /**
+   * The actions at the bottom, grouped in a ButtonGroup
+   */
+  actions: ReactNode;
+
+  /**
+   * The text to display in the modal body
+   */
+  bodyText: ReactNode | ReactNode[];
```
gregaubert, line 35: `ReactNode` definition contains `Iterable<ReactNode>` so that should cover the `ReactNode[]` already no?

Or do you have a specific use case in mind for `ReactNode[]`?

stories/Spotlight-stories.tsx
```diff
@@ -0,0 +1,110 @@
...
+        target: '#step4',
+      },
+      {
+        bodyText: "The fifth step's description",
+        headerText: 'Fifth and final step',
+        placement: SpotlightModalPlacement.Center,
+        target: '#step5',
+      },
+    ],
+  },
```
gregaubert, line 73: Let's move the `args` definition in the story directly like we do everywhere else.

src/components/spotlight/SpotlightTypes.tsx
```diff
@@ -0,0 +1,123 @@
...
+   * Shoud we disable closing the spotlight when clicking the overlay?
+   * Defaults to false when there's only one step, true otherwise (optional)
+   */
+  shouldDisableOverlayClose?: boolean;
+
+  /**
+   * Should we display the tour progress in the Next button in continuous tours?
+   * Defaults to false (optional)
+   */
+  shouldShowProgressInNextButton?: boolean;
```
gregaubert, line 97: In the story that doesn't seem to work? Maybe something is missing in the story?
Is this used in the products?

src/components/spotlight/SpotlightTypes.tsx
```diff
@@ -0,0 +1,123 @@
...
+
+  /**
+   * A function called when the Spotlight state changes (optional) - see https://docs.react-joyride.com/props
+   */
+  callback?: JoyrideProps['callback'];
+
+  /**
+   * The label for the Close button - defaults to "Close" (optional)
+   */
+  closeLabel?: string;
```
gregaubert, line 65: In the story this doesn't seem to have any effect?

stories/Spotlight-stories.tsx
```diff
@@ -0,0 +1,110 @@
...
+
+  component: Spotlight,
+  decorators: [basicWrapperDecorator],
+
+  parameters: {
+    controls: {
+      exclude: ['activeTargetClassName', 'callback', 'image'],
```
gregaubert, line 80: Maybe use the `toDisabledControlArgType` instead to style have them visible and "documented" in the story?

gregaubert (commented):
A few more comments, mainly nitpicking but also a few things that might not be working in the story?

src/components/spotlight/SpotlightTypes.tsx
```diff
@@ -0,0 +1,123 @@
...
+
+  /**
+   * The image to display at the top of each spotlight modal (optional)
+   */
+  image?: ReactNode;
+
+  /**
+   * Should the tour be played sequentially with the Next button? - defaults to false (optional)
+   */
+  isContinuous?: boolean;
```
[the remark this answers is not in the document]
david-cho-lerat-sonarsource: that's right because we decided to handle this ourselves: 1 step is not continuous, >1 is continuous. So we don't didn't do anything with this prop, I've removed it.

src/components/spotlight/SpotlightTypes.tsx
david-cho-lerat-sonarsource: it's actually useless, it was only mentioned once in the products and the condition setting it to true was never triggered (might have been in the past). I'm removing it.

src/components/spotlight/SpotlightTypes.tsx
```diff
@@ -0,0 +1,126 @@
...
+ * along with this program; if not, write to the Free Software Foundation,
+ * Inc., 51 Franklin Street, Fifth Floor, Boston, MA  02110-1301, USA.
+ */
+
+import { ReactNode } from 'react';
+import {
+  Callback as JoyrideCallback,
+  CallBackProps as SpotlightCallbackProps,
+} from 'react-joyride';
+import { SpotlightModalProps } from './SpotlightModal';
```
jeremy-davis-sonarsource, line 26: NP: I'd move the `SpotlightModalProps` in here: I wouldn't expect a `types` module to import from a component module, but rather the opposite.

PR #387, opened 2025-06-23
ECHOES-788 Implement SearchInput component

gregaubert:


src/components/toggle-button-group/ToggleButtonGroup.tsx
```diff
@@ -102,6 +102,8 @@ function ToggleButtonItem(props: ToggleButtonItemProps) {
 }
 
 const StyledRoot = styled(RadixToggleGroup.Root)`
+  isolation: isolate; // Prevents the z-index changes of the items inside from affecting layers outside the ToggleButtonGroup
```
jeremy-davis-sonarsource, line 105: The comment is a bit vague, what specific z-index conflict is this solving?

jeremy-davis-sonarsource (approved):
Nice! just a few minor things

src/components/spinner/Spinner.tsx
```diff
@@ -52,19 +53,22 @@ export const Spinner = forwardRef<HTMLSpanElement, SpinnerProps>((props, ref) =>
...
     <>
-      <SpinnerWrapper className={wrapperClassName} inline={children === undefined}>
+      <SpinnerWrapper className={wrapperClassName} inline={isInline}>
         <SpinnerInner
           {...radixProps}
           aria-live="polite"
           className={classNames({
             it__loading: isLoading,
           })}
+          inline={isInline || isDefined(label)}
```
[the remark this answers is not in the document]
gregaubert: Yeah it is, mainly because the wrapper has to align with the outside.
While the SpinnerInner needs to change depending if there is a visible label or not, if yes it needs to be inline so that it displays correctly next to the label.

src/components/toggle-button-group/ToggleButtonGroup.tsx
gregaubert: Updated the comment to be more specific.

But I think it's juste a general rule of thumb, if we use z-index in one of our component to change the layer order of that component rendering, these layer order changes must be scoped to that component only, and we can achieve that by creating a new stacking context at the root of the component.

Non of our fixed z-index in the library should have effect in the outside world.

PR #398, opened 2025-06-26
ECHOES-796 Rename "type" prop to "variety" for MessageCallout, MessageInline

david-cho-lerat-sonarsource:
[ECHOES-796](https://sonarsource.atlassian.net/browse/ECHOES-796)

[ECHOES-796]: https://sonarsource.atlassian.net/browse/ECHOES-796?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/components/form/FormField.tsx
+++ b/src/components/form/FormField.tsx
@@ -20,7 +20,7 @@
 import styled from '@emotion/styled';
 import { type ComponentProps, type JSX, forwardRef } from 'react';
 import { TextNodeOptional } from '~types/utils';
-import { MessageInline, MessageInlineSize, MessageType } from '../messages';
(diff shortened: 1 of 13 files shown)


gregaubert:
Ok actually that's fine, we just need to do it for Banner too then

gregaubert:
Also for the tickets were we have breaking changes like this, I suggest we directly create a matching PR in sonarcloud-webapp with the updates. We will consolidates these into a common branch into sonarcloud-webapp repo as we progress.

And second thing, we should create a draft release for v1 where we document all these breaking change as we introduce them.

PR #416, opened 2025-07-07
ECHOES-819 Bump version to 1.0.0 in echoes-v1-common-branch

jeremy-davis-sonarsource:


src/components/modals/ModalBody.tsx
gregaubert, line 1: I see now that we are not throttling or anything, should we?

gregaubert (approved):
Looks great!

src/components/modals/ModalBody.tsx
jeremy-davis-sonarsource: As discussed, probably not important

PR #440, opened 2025-07-29
ECHOES-875 Add missing base Tailwind height and width in tailwindConfig

gregaubert:
[ECHOES-875](https://sonarsource.atlassian.net/browse/ECHOES-875)

Part of 

[ECHOES-875]: https://sonarsource.atlassian.net/browse/ECHOES-875?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/design-tokens/build.js
+++ b/design-tokens/build.js
@@ -255,7 +255,10 @@ function buildTailwindConfig() {
   const height = mapTokens(jsonTokens, 'height');
 
   addCoreTokens(height);
+  height.screen = '100vh';
+
   addCoreTokens(width);
+  width.screen = '100vw';
 
   const result = {
     echoesPreset: {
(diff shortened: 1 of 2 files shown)


design-tokens/build.js
```diff
@@ -298,6 +301,14 @@ function addCoreTokens(tokens) {
   tokens.min = 'min-content';
   tokens.max = 'max-content';
 
+  tokens.px = '1px';
+  tokens.dvw = '100dvw';
+  tokens.dvh = '100dvh';
+  tokens.lvw = '100lvw';
+  tokens.lvh = '100lvh';
+  tokens.svw = '100svw';
+  tokens.svh = '100svh';
```
jeremy-davis-sonarsource, line 310: These should be split as well: dvw is for width, while dvh is for height.

jeremy-davis-sonarsource (changes requested):

design-tokens/build.js
gregaubert: Looking at tailwind documentation: https://tailwindcss.com/docs/height
They provide both for height and for width, that's why I did the same. I really think we should just be aligned with the default on these things.

jeremy-davis-sonarsource (approved):
I'll let you decide

PR #453, opened 2025-08-12
ECHOES-893 Fix Spinner layout

jeremy-davis-sonarsource:


jeremy-davis-sonarsource:
> It looks good but I've got some weird rendering in some components, on the buttons for example:
> 
Wow, I was so focused on the layout thing, I didn't think any other Spinner would be impacted :see_no_evil: 
It's the `SpinnerOverrideColor` that's not playing well with the move of  `className`, of course...

gregaubert:
> > It looks good but I've got some weird rendering in some components, on the buttons for example:
> 
> Wow, I was so focused on the layout thing, I didn't think any other Spinner would be impacted 🙈 It's the `SpinnerOverrideColor` that's not playing well with the move of `className`, of course...

Ah yeah that make sense

gregaubert (approved):

PR #467, opened 2025-08-26
ECHOES-901 Add PageHeader

david-cho-lerat-sonarsource:
[ECHOES-901](https://sonarsource.atlassian.net/browse/ECHOES-901)

[ECHOES-901]: https://sonarsource.atlassian.net/browse/ECHOES-901?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/components/page-header/PageHeader.tsx
+++ b/src/components/page-header/PageHeader.tsx
@@ -1,0 +1,163 @@
+/*
+ * Echoes React
+ * Copyright (C) 2023-2025 SonarSource SA
+ * mailto:info AT sonarsource DOT com
+ *
+ * This program is free software; you can redistribute it and/or
+ * modify it under the terms of the GNU Lesser General Public
+ * License as published by the Free Software Foundation; either
+ * version 3 of the License, or (at your option) any later version.
+ *
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the GNU
+ * Lesser General Public License for more details.
+ *
+ * You should have received a copy of the GNU Lesser General Public License
+ * along with this program; if not, write to the Free Software Foundation,
+ * Inc., 51 Franklin Street, Fifth Floor, Boston, MA  02110-1301, USA.
+ */
+
+import styled from '@emotion/styled';
+import { forwardRef, ReactNode } from 'react';
+
+import { cssVar } from '~utils/design-tokens';
+import { Divider } from '../divider';
+
+export interface PageHeaderProps {
+  /**
+   * Action elements (e.g. a set of <Button>s) to display on the right side of the header.
+   * Use <PageHeader.Actions> to wrap them.
+   */
+  actions?: ReactNode;
+  /**
+   * Breadcrumb navigation elements to display above the title.
+   * Use <PageHeader.Breadcrumbs> to wrap them.
+   */
+  breadcrumbs?: ReactNode;
+  /**
+   * Additional CSS class name(s)
+   */
+  className?: string;
+  /**
+   * Description text to display below the title. Use <PageHeader.Description> to wrap it.
+   */
+  description?: ReactNode;
+  /**
+   * Whether to show a divider line at the bottom of the header
+   * @default false
+   */
+  hasDivider?: boolean;
+  /**
+   * Metadata elements to display below the title. Use <PageHeader.Metadata> to wrap them.
+   */
+  metadata?: ReactNode;
+  /**
+   * Navigation elements to display at the bottom. Use <PageHeader.Navigation> to wrap them.
+   */
+  navigation?: ReactNode;
+  /**
+   * The main title content (required). Use <PageHeader.Title> to wrap it.
+   */
+  title: ReactNode;
+}
+
+/**
+ * A flexible page header component that displays a title, optional metadata, description,
+ * breadcrumbs, and actions. Can optionally include a divider, and <PageHeader.Navigation> elements
+ * as children.
+ */
+export const PageHeaderRoot = forwardRef<HTMLDivElement, PageHeaderProps>((props, ref) => {
+  const { actions, breadcrumbs, description, hasDivider, metadata, navigation, title, ...rest } =
+    props;
+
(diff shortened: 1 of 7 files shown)


src/components/page-header/PageHeaderMetadata.tsx
```diff
@@ -0,0 +1,51 @@
...
+  return (
+    <StyledPageHeaderMetadata ref={ref} {...rest}>
+      {typeof children === 'string' ? <Text isSubtle>{children}</Text> : children}
+    </StyledPageHeaderMetadata>
+  );
+});
+
+PageHeaderMetadata.displayName = 'PageHeaderMetadata';
+
+const StyledPageHeaderMetadata = styled.div`
```
gregaubert, line 46: Let's not forget to add the `StyledComponents.displayName` to this styled component

src/components/page-header/PageHeaderTitle.tsx
```diff
@@ -0,0 +1,56 @@
...
+
+        {suffix}
+      </StyledPageHeaderTitle>
+    );
+  },
+);
+
+PageHeaderTitle.displayName = 'PageHeaderTitle';
+
+const StyledPageHeaderTitle = styled.div`
```
gregaubert, line 51: Let's not forget to add the `StyledComponents.displayName` to this styled component

src/components/page-header/PageHeader.tsx
gregaubert, line 1: I think we are missing some handling for long texts in the title, description, metadata, should they wrap? Or ellipsis?

src/components/page-header/PageHeader.tsx
```diff
@@ -0,0 +1,120 @@
...
+
+            {actions && <StyledPageHeaderMainRight>{actions}</StyledPageHeaderMainRight>}
+          </StyledPageHeaderMain>
+        </StyledPageHeaderTop>
+
+        {(hasDivider || children) && (
+          <StyledPageHeaderBottom>
+            {children}
+
+            <Divider />
```
jeremy-davis-sonarsource, line 71: This makes the border appear below the menu, which is different than the design.

image
image

It seems the GlobalNav also has this behavior, so let's double check with Marcio!

src/components/page-header/PageHeader.tsx
```diff
@@ -0,0 +1,120 @@
...
+  flex-direction: column;
+  gap: ${cssVar('dimension-space-200')};
+`;
+
+const StyledPageHeaderBottom = styled.div`
+  white-space: nowrap;
+  width: 100%;
+`;
+
+const StyledPageHeaderMain = styled.div`
```
gregaubert, line 102: Should we add a bit of gap to make sure left and right will never touch?

src/components/page-header/PageHeader.tsx
```diff
@@ -0,0 +1,120 @@
...
+
+export interface PageHeaderProps {
+  actions?: ReactNode;
+  breadcrumbs?: ReactNode;
+  className?: string;
+  description?: ReactNode;
+  hasDivider?: boolean;
+  metadata?: ReactNode;
+  title: ReactNode;
+}
```
jeremy-davis-sonarsource, line 38: tsdoc!

stories/PageHeader-stories.tsx
```diff
@@ -0,0 +1,120 @@
...
+        }>
+        Page title
+      </PageHeader.Title>
+    ),
+  },
+  parameters: {
+    controls: {
+      exclude: ['actions', 'breadcrumbs', 'children', 'description', 'metadata', 'title'],
+    },
+  },
```
gregaubert, line 111: It's weird to have 0 controls on the story, maybe there is a better way to provide customization in the story? Using select type controls to toggle or not some of the children?

jeremy-davis-sonarsource (changes requested):
Looks good!
We need the tsdoc to explain the API, cause it's definitely not trivial. 

I'll ping Marcio about the border alignment

src/components/page-header/PageHeader.tsx
david-cho-lerat-sonarsource: For now they wrap, which I think is OK by default for a page header. The metadata is a different animal: if it's composed of a text and badges, like in Figma, the text should probably have ellipses, but I don't think we should handle this here? Marcio said the metadata should just be a "slot"

src/components/page-header/PageHeader.tsx
david-cho-lerat-sonarsource: this isn't in the Figma design, but I guess a 100 gap (8px) can't hurt? I'll add it in

src/components/page-header/PageHeader.tsx
david-cho-lerat-sonarsource: as @jeremy-davis-sonarsource suggested, I've used a negative margin instead

src/components/page-header/PageHeader.tsx
gregaubert: Looking at the title it doesn't seems like it's correctly handled right now, I'm seeing this:

Image

david-cho-lerat-sonarsource: oh yes sorry, I've removed the nowrap

src/components/page-header/PageHeader.tsx
```diff
@@ -0,0 +1,161 @@
...
+          </StyledPageHeaderMain>
+        </StyledPageHeaderTop>
+
+        {(hasDivider || children) && (
+          <StyledPageHeaderBottom>
+            {children}
+
+            {children ? <StyledDividerWithOverlap /> : <Divider />}
+          </StyledPageHeaderBottom>
+        )}
```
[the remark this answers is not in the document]
david-cho-lerat-sonarsource: Idk I was thinking about title as well but then as it's mandatory it didn't quite fit as "children", which can be empty..
It felt weird having no "children" and all props, and I thought that this navigation "page chin" was really what stood out from the rest, as not really being part of the page header, and it's optional, so why not use that as children? But I'm not 100% convinced either. @jeremy-davis-sonarsource WDYT?

jeremy-davis-sonarsource (approved):

PR #471, opened 2025-08-28
ECHOES-919 Add layout sidebar navigation collapse behavior

gregaubert:


src/components/sidebar-navigation/SidebarNavigation.tsx
```diff
@@ -49,35 +53,57 @@ export const SidebarNavigation = forwardRef<
...
+  --sidebar-navigation-width: ${cssVar('sidebar-navigation-sizes-width-open')};
+
+  // hover and focus-within pilotes the open state of the sidebar
+  [data-sidebar-docked='false'] &:not(:hover, :focus-within) {
+    --sidebar-navigation-width: ${cssVar('sidebar-navigation-sizes-width-closed')};
+  }
+
+  [data-sidebar-docked='true'] &:is(:hover, :focus-within) {
+    box-shadow: ${cssVar('box-shadow-x-large')};
+  }
```
jeremy-davis-sonarsource, line 107: should be `data-sidebar-docked='false'`

src/components/sidebar-navigation/SidebarNavigationAccordionItem.tsx
```diff
@@ -142,18 +142,19 @@ const AccordionItemPanel = styled.section`
...
   border-left: ${cssVar('border-width-default')} solid ${cssVar('color-border-weak')};
 
-  [data-sidebar-collapsed='true'] & {
+  // The children SidebarNavigationItems rely on this css property to set their display value, falling back to flex if not inside an accordion
+  --sidebar-navigation-accordion-children-display: flex;
+
+  // We force the children SidebarNavigationItems to be hidden when the accordion is closed or the sidebar is closed
+  &[data-accordion-open='false'] {
+    --sidebar-navigation-accordion-children-display: none;
+  }
```
jeremy-davis-sonarsource, line 151: The comment is a bit confusing. How about
`The next two rules hide the child SidebarNavigationItems when the accordion is closed, and also when the sidebar is neither docked nor open (hovered or focused).`

src/components/sidebar-navigation/SidebarNavigationItem.tsx
```diff
@@ -44,11 +44,11 @@ export interface SidebarNavigationItemProps
   children: TextNode;
   className?: string;
   /**
-   * Whether to hide the Icon when the sidebar is expanded.
-   * The purpose is to have the icon appear only when the sidebar is collapsed,
+   * Whether to hide the Icon when the sidebar is open.
+   * The purpose is to have the icon appear only when the sidebar is open,
    * and for accordion child items only.
```
jeremy-davis-sonarsource, line 49: Oops, no, the opposite! `only when the sidebar is neither docked nor open`

jeremy-davis-sonarsource (approved):
I'm wondering about the focus behavior, otherwise it looks great!
Just a few fixes about comments and the reversed condition for the shadow to switch

src/components/sidebar-navigation/SidebarNavigation.tsx
gregaubert: Indeed, I caught it while working on some tests and fixed with my last commit

PR #476, opened 2025-09-02
@gregaubert ECHOES-924 Cleanup layout components export structure

gregaubert:
ECHOES-924


src/components/layout/index.ts
```diff
@@ -54,15 +75,24 @@ export type { AsideProps, PageGridProps } from './LayoutSlots';
...
 export const Layout = Object.assign(LayoutRoot, {
+  /**
+   * The left aside container, that fills the Aside area of the ContentGrid.
+   * Should be used for filters, facets or navigation lists.
+   */
   AsideLeft,
+
+  /**
+   * The banner container, that holds all the product banners and fills BannerContainer area of the
+   * Layout grid. It can eventually hold multiple banners but ideally should only hold one at a time.
```
david-cho-lerat-sonarsource, line 89:
```suggestion
   * Layout grid. It can optionally hold multiple banners but ideally should only hold one at a time.
```

gregaubert: Fixed in the next PR xD

PR #507, opened 2025-09-26
ECHOES-957 Expose header level prop in PageHeader.Title

jeremy-davis-sonarsource:
[ECHOES-957](https://sonarsource.atlassian.net/browse/ECHOES-957)

[ECHOES-957]: https://sonarsource.atlassian.net/browse/ECHOES-957?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/components/layout/page-header/PageHeaderTitle.tsx
+++ b/src/components/layout/page-header/PageHeaderTitle.tsx
@@ -19,46 +19,70 @@
  */
 
 import styled from '@emotion/styled';
-import { forwardRef, PropsWithChildren, ReactNode } from 'react';
+import { ForwardedRef, forwardRef, PropsWithChildren, ReactNode } from 'react';
 import { cssVar } from '~utils/design-tokens';
-import { Heading } from '../../typography';
+import { Heading, HeadingProps } from '../../typography';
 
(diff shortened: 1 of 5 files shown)


src/components/layout/page-header/PageHeaderTitle.tsx
```diff
@@ -19,47 +19,71 @@
...
+   * The heading tag to use for the title itself
+   */
+  headingLevel: HeadingProps['as'];
+}
+
+function renderFunction(
+  props: PropsWithChildren<PageHeaderTitleProps>,
+  ref: ForwardedRef<HTMLDivElement>,
+) {
+  const { children, headingLevel, prefix, suffix, ...rest } = props;
```
david-cho-lerat-sonarsource, line 55: so, we don't want to use a default for headingLevel? Is this to force user to think about the actual level that's needed in the context? But won't that lead to having to write `headingLevel="h1"` in most instances, except when we have a ContextHeader? Wouldn't that qualify as a reasonable default?

PR #673, opened 2026-04-24
ECHOES-1015 New MarginIndicator component

jeremy-davis-sonarsource:
Integration PR: [sonarcloud-webapp#5334](https://github.com/SonarSource/sonarcloud-webapp/pull/5334)


src/components/margin-indicator/MarginIndicator.tsx
```diff
@@ -0,0 +1,147 @@
...
+  border-radius: unset;
+  padding: 0;
+  height: ${MARGIN_INDICATOR_HEIGHT}px;
+  width: ${MARGIN_INDICATOR_WIDTH}px;
+
+  &:focus-visible {
+    border-radius: ${cssVar('border-radius-400')};
+
+    position: relative;
+    z-index: 1;
```
gregaubert, line 148: The z-index only applies to the focus ring there? Or do we set the z-index only when the button is in focus state? Does it always make sense to apply this z-index?

src/components/margin-indicator/MarginIndicator.tsx
```diff
@@ -0,0 +1,147 @@
...
+  ),
+};
+
+function MarginIndicatorSvg({
+  ariaLabel,
+  indicatorType,
+  svgRef,
+  ...radixProps
+}: Readonly<{
+  ariaLabel?: string;
```
gregaubert: Should we have a default aria-label based on the type?

src/components/margin-indicator/MarginIndicator.tsx
```diff
@@ -0,0 +1,147 @@
...
+  indicatorType: `${MarginIndicatorType}`;
+  svgRef?: React.Ref<SVGSVGElement>;
```
gregaubert: can we just call it `ref` now with react 19 changes to refs?

src/components/margin-indicator/MarginIndicator.tsx
```diff
@@ -0,0 +1,147 @@
...
+      viewBox={`0 0 ${MARGIN_INDICATOR_WIDTH} ${MARGIN_INDICATOR_HEIGHT}`}
+      width={MARGIN_INDICATOR_WIDTH}
+      xmlns="http://www.w3.org/2000/svg">
+      {ariaLabel && <title>{ariaLabel}</title>}
+      {PATTERN[indicatorType]}
+    </svg>
+  );
+}
+
+export const MarginIndicator = forwardRef<SVGSVGElement | HTMLButtonElement, MarginIndicatorProps>(
```
gregaubert: No need to use forwardRef anymore with React 19 right? 
https://react.dev/reference/react/forwardRef

src/components/margin-indicator/MarginIndicator.tsx
```diff
@@ -0,0 +1,147 @@
...
+        indicatorType={indicatorType}
+        svgRef={ref as React.Ref<SVGSVGElement>}
+        {...radixProps}
+      />
+    );
+  },
+);
+MarginIndicator.displayName = 'MarginIndicator';
+
+const IndicatorButton = styled(ButtonStyled)`
```
gregaubert, line 136: Let's add a displayName for it

src/components/margin-indicator/MarginIndicator.tsx
```diff
@@ -0,0 +1,147 @@
...
+        fill={cssVar('margin-indicator-colors-background-uncovered')}
+        height="20"
+        width="2"
+        x="4"
+      />
+    </>
+  ),
+};
+
+function MarginIndicatorSvg({
```
gregaubert, line 87: Let's add a displayName for it

gregaubert (commented):
A few questions, but looks good overall!

src/components/margin-indicator/MarginIndicator.tsx
jeremy-davis-sonarsource: it applies to the whole thing (you can't apply it only to a part)
It always makes sense, otherwise the ring appears beneath the `LineCode`, so it looks like `C`
(and also beneath the following MarginIndicator, if there is one)

src/components/margin-indicator/MarginIndicator.tsx
jeremy-davis-sonarsource: I considered it and decided it was too involved, but no strong opinion

gregaubert (approved):
Noice!

PR #681, opened 2026-05-07
ECHOES-1305 Add script to add material icons

jeremy-davis-sonarsource:


config/scripts/create-icon.js
```diff
@@ -0,0 +1,171 @@
+/*
```
david-cho-lerat-sonarsource, line 1: NP: no "filled" support? maybe not a big deal

config/scripts/create-icon.js
```diff
@@ -0,0 +1,171 @@
...
+ * Examples:
+ *   node config/scripts/create-icon.js Shield shield
+ *   node config/scripts/create-icon.js Shield shield e9e0
+ */
+
+import { execSync } from 'node:child_process';
+import fs from 'node:fs';
+import { join } from 'node:path';
+
+const CODEPOINTS_URL =
```
david-cho-lerat-sonarsource, line 44: NP: this may contain newer codepoints than our local font includes. So the component will look OK, but not actually render. We could check if we have the requested codepoint with `subset-font` maybe?

david-cho-lerat-sonarsource (approved):
LGTM, 2 NPs on the script 💪

config/scripts/create-icon.js
jeremy-davis-sonarsource: good point!

config/scripts/create-icon.js
jeremy-davis-sonarsource: good idea, should be simple

PR #687, opened 2026-05-12
ECHOES-1311 Pass brand value to build step in unified dogfooding workflow

david-cho-lerat-sonarsource:
The build step in the unified dofgooding workflow is missing the now-mandatory brand value.
Also, the sonar-s-public runners have been deprecated in favor of sonar-s, which can now handle public repositories.

A run of the dogfooding workflow on this PR's branch shows that it works: https://github.com/SonarSource/echoes-react/actions/runs/25716453852

### DIFF
--- a/.github/workflows/unified-dogfooding.yml
+++ b/.github/workflows/unified-dogfooding.yml
@@ -7,7 +7,7 @@ on:
 
 jobs:
   shadow-scans:
-    runs-on: sonar-s-public
+    runs-on: sonar-s
     name: Build and run shadow scans
     permissions:
       id-token: write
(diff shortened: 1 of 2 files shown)


gregaubert (approved):

.github/workflows/unified-dogfooding.yml
```diff
@@ -20,6 +20,8 @@ jobs:
       - uses: SonarSource/ci-github-actions/build-yarn@v1
         with:
           run-shadow-scans: true
+        env:
+          DESIGN_TOKEN_BRAND: Brand-A
```
gregaubert, line 24: We will need to find a better name than "Brand-A" xD

PR #693, opened 2026-06-04
ECHOES-1357 Aggregate repeated toasts with a counter badge

david-cho-lerat-sonarsource:
## Summary

- Aggregate repeated toast calls when they have the same plain-text `description`, the same `variety`, and no explicit `id`.
- If the toast also has a visible plain-text `title`, that title must match too. Description-only toasts aggregate too, and blank plain-text titles are treated like no title.
- Keep a single visible toast for those repeats, keep the original text unchanged, and show a counter badge inline next to the title, or next to the description when there is no title.
- Cap the visible badge at `99+` while keeping the full count for screen readers.
- Clear the internal aggregation tracking when the aggregated toast is dismissed or auto-closed, so a later matching toast starts again from `1`.

## Notes

- Reusing an explicit `id` still follows the existing stable-id update behavior and bypasses automatic same-text aggregation.
- Toasts with actions, lifecycle callbacks, non-text descriptions, or non-text titles are not automatically aggregated.
- Storybook includes both the titled and description-only aggregated examples.
- The public dismiss API remains `toast.dismiss(...)`.

### DIFF
--- a/src/common/components/Toast.tsx
+++ b/src/common/components/Toast.tsx
@@ -17,6 +17,7 @@
  * along with this program; if not, write to the Free Software Foundation,
  * Inc., 51 Franklin Street, Fifth Floor, Boston, MA  02110-1301, USA.
  */
+
 import { css, Global } from '@emotion/react';
 import styled from '@emotion/styled';
 import { forwardRef, ReactNode, useCallback, useMemo } from 'react';
@@ -24,7 +25,7 @@ import { useIntl } from 'react-intl';
 import { toast as sonnerToast } from 'sonner';
 import { TextNode, TextNodeOptional } from '~types/utils';
 import { cssVar } from '~utils/design-tokens';
-import { ButtonIcon, Spinner, Text } from '../../components';
+import { BadgeCounter, ButtonIcon, Spinner, Text } from '../../components';
 import { IconCheckCircle, IconError, IconInfo, IconWarning, IconX } from '../../components/icons';
 import { ScreenReaderPrefix } from './ScreenReaderPrefix';
 
@@ -68,7 +69,8 @@ interface ToastActionsParams {
    */
   id: ToastId;
   /**
-   * Function to programmatically dismiss the toast, after clicking an action we should normally also dismiss the toast.
+   * Function to programmatically dismiss the toast. After clicking an action,
+   * we should normally also dismiss the toast.
    */
   dismiss: VoidFunction;
 }
@@ -77,7 +79,8 @@ export interface ToastProps {
   /**
    * Custom actions to display in the toast (optional). Receives the toast ID
    * and dismiss function as parameters. It should be either one or more Buttons or Links.
-   * If provided, the toast should also have the `isDismissable` prop set to true and it's duration set to infinite.
+   * If provided, the toast should also have the `isDismissable` prop set to true
+   * and its duration set to infinite.
    */
   actions?: ({ id, dismiss }: ToastActionsParams) => ReactNode;
   /**
@@ -95,6 +98,11 @@ export interface ToastProps {
    * The default is false.
    */
   isDismissable?: boolean;
+  /**
+   * Number of times the same toast has been repeated. When greater than 1, a counter badge is
+   * displayed next to the toast.
+   */
+  repetitionCount?: number;
   /**
    * Optional prefix text for screen readers, providing additional context.
    * If not provided, a default message based on the toast variety will be used.
@@ -142,12 +150,39 @@ export const Toast = forwardRef<HTMLDivElement, Readonly<ToastProps>>((props, re
     description,
     id,
     isDismissable = false,
+    repetitionCount: repetitionCountProp,
     screenReaderPrefix,
     title,
     variety,
     ...htmlProps
   } = props;
+
   const intl = useIntl();
+  const repetitionCount = repetitionCountProp ?? 0;
+  const hasTitleLine = hasToastTitleLine(title);
+
+  // Keep the visible badge compact while screen readers still hear the full count.
+  const repetitionBadgeValue = repetitionCount > 99 ? '99+' : repetitionCount;
+
+  const repetitionCounter =
+    repetitionCount > 1 ? (
+      <>
+        <ScreenReaderPrefix>
+          {intl.formatMessage(
+            {
+              id: 'toast.repetition-count',
+              defaultMessage: 'Shown {count} times',
+              description: 'Screen reader label for a toast shown multiple times',
+            },
+            { count: repetitionCount },
+          )}
+        </ScreenReaderPrefix>
+
+        <ToastRepetitionCounter aria-hidden>
+          <BadgeCounter value={repetitionBadgeValue} />
+        </ToastRepetitionCounter>
+      </>
+    ) : undefined;
 
   const handleDismiss = useCallback(() => {
     sonnerToast.dismiss(id);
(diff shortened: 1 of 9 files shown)


src/components/badges/BadgeCounter.tsx
```diff
@@ -19,11 +19,10 @@
  */
 
 import styled from '@emotion/styled';
-import { forwardRef } from 'react';
+import { ComponentProps, forwardRef } from 'react';
 import { cssVar } from '~utils/design-tokens';
 
-export interface BadgeCounterProps {
-  className?: string;
+export interface BadgeCounterProps extends Omit<ComponentProps<'span'>, 'children'> {
```
david-cho-lerat-sonarsource: the type was not reflecting what the component was actually accepting. Now we're using it with `aria-hidden` explicitly, this type is more faithful to the truth.

src/utils/toasts.tsx
```diff
@@ -241,6 +311,378 @@ const TOAST_DURATION_MAP = {
...
+  if (typeof node === 'string') {
+    return node;
+  }
+
+  if (isValidElement(node) || !isIterable(node)) {
+    return undefined;
+  }
+
+  let text = '';
+
```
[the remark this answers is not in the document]
david-cho-lerat-sonarsource: Not applying this change.

`title` and `description` are typed through `TextNode` / `TextNodeOptional`, and those exported types currently support strings, JSX elements, and iterables of those values, but not numbers. The aggregation logic in `getPlainText()` is intentionally aligned with that public contract.

React may render numeric children at runtime, but supporting numeric toast content for aggregation would be an API change, not a local fix. If we want that behavior, it should be done explicitly by widening the exported types and documenting it, rather than silently extending support here.

src/utils/toasts.tsx
```diff
@@ -165,6 +213,15 @@ type ToastFn = {
  * Calling `toast` or one of the variety shortcuts again with that `id` updates the existing toast
  * instead of creating a second one.
  *
+ * **Aggregating repeated toasts**
+ *
+ * When no explicit `id` is provided, repeated calls with the same plain-text `title`, the same
+ * plain-text `description`, and the same `variety` reuse the existing toast instead of creating
+ * duplicates. The toast keeps its original text and shows a counter badge on the right.
```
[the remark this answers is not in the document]
david-cho-lerat-sonarsource: This remark is valid. The PR description was stale.

The current implementation does not introduce `registerToastDismissHandler` or `dismissToast`, and it does not add a custom timer-refresh layer. The public API remains `toast.dismiss(...)`, and aggregation cleanup is handled internally through the Sonner `onAutoClose` / `onDismiss` path.

I updated the PR description so it matches the code as it stands now.

stories/Toast-stories.tsx
```diff
@@ -184,3 +193,60 @@ export const StableIdUpdates: Story = {
...
+      },
+    },
+  },
+  render: (args) => {
+    const {
+      description = 'File uploaded',
+      isDismissable = true,
+      title = 'Upload complete',
+      ...restArgs
+    } = args;
```
jeremy-davis-sonarsource: It seems stroybook injects a function automatically, so it breaks the tracking (since onDismiss is defined)

stories/Toast-stories.tsx
```diff
@@ -184,3 +193,60 @@ export const StableIdUpdates: Story = {
...
+
+    return (
+      <Button
+        onClick={() => {
+          toast.success({
```
jeremy-davis-sonarsource: We should call `toast()`, to preserve the chosen `variety`

src/utils/__tests__/toasts-test.tsx
```diff
@@ -18,47 +18,49 @@
...
+  afterEach(resetToastTestState);
 
   it('should create and render a toast with required parameters', async () => {
     render(<div />);
 
-    toast({
-      variety: ToastVariety.Info,
-      description: TEST_MESSAGE,
-    });
+    trackToastId(
```
jeremy-davis-sonarsource, line 42: Why track the toast id?

jeremy-davis-sonarsource (changes requested):
Why don't we allow aggregation when onDismiss or onAutoClose is defined?

Looks good, apart from the stories not working properly

stories/Toast-stories.tsx
david-cho-lerat-sonarsource: yeah I had realized the stories were broken and have fixed them

stories/Toast-stories.tsx
david-cho-lerat-sonarsource: Good point. Fixed.

src/utils/__tests__/toasts-test.tsx
david-cho-lerat-sonarsource: The tests track every created toast id so afterEach can dismiss leftovers deterministically and wait for Sonner’s deferred removal. That keeps the toast tests isolated and avoids state leaking between timer-based cases.

david-cho-lerat-sonarsource:
> Why don't we allow aggregation when onDismiss or onAutoClose is defined?
> 
I kept lifecycle callbacks as an explicit opt-out because aggregation collapses several toast calls into one visible toast lifecycle. With callbacks, the behavior becomes ambiguous: once per underlying call, once per visible aggregated toast, or only for the last update..? I preferred the explicit opt-out over inventing callback semantics in this PR.

src/utils/toast-internals/repeated-toast-tracking.ts
```diff
@@ -0,0 +1,158 @@
...
+    isDefined(params.onDismiss)
+  ) {
+    return undefined;
+  }
+
+  // Automatic aggregation requires a visible plain-text description. When a title is present, it
+  // must also be visible plain text and becomes part of the aggregation key.
+  const hasTitle = isDefined(params.title) && params.title !== false;
+  const titleText = hasTitle ? getToastPlainText(params.title)?.trim() : undefined;
+  const descriptionText = getToastPlainText(params.description)?.trim();
```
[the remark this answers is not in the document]
david-cho-lerat-sonarsource: Documented this in the public TSDoc instead of widening the aggregation key. The intended rule for this PR stays aligned with the UX behavior: automatic aggregation keys on visible text plus \, and on the visible plain-text \ when one is shown. I added a note in \ clarifying that when automatic aggregation happens, non-key props such as \, \, \, and \ come from the latest matching call and are applied to the shared toast.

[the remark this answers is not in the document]
david-cho-lerat-sonarsource: Kept the split, but made the intent clearer in code instead of extracting a shared helper. The render-side helper in `Toast.tsx` is now `hasToastTitleLine(...)`, while `repeated-toast-tracking.ts` keeps the plain-text-only matching logic for aggregation. I also added comments in both places to make the separation explicit: rendering decides whether the toast shows a title line at all, while aggregation decides whether a title can participate in automatic same-text matching. Because the repetition badge is only shown once aggregation has already succeeded, this split does not currently create a user-visible inconsistency.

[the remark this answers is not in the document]
david-cho-lerat-sonarsource: that's by design

src/common/components/Toast.tsx
```diff
@@ -217,13 +274,51 @@ const ToastContent = styled.div`
...
+  if (typeof title === 'string') {
+    return title.trim().length > 0;
+  }
+
+  if (Array.isArray(title)) {
+    return title.some((child) => hasVisibleToastTitle(child));
+  }
+
+  return Boolean(title);
+}
```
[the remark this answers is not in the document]
david-cho-lerat-sonarsource: The split here is intentional, so I kept the behavior and clarified the naming/comments instead of extracting a shared helper.

Toast.tsx is deciding whether a title line should be rendered at all, which is broader than aggregation eligibility. repeated-toast-tracking.ts is deciding whether a title can participate in automatic same-text aggregation, which is intentionally narrower and only accepts visible plain-text titles. Because the repetition badge is only shown when aggregation has already succeeded, this difference does not currently produce a user-visible inconsistency.

I renamed the render-side helper to shouldRenderToastTitleLine, renamed the aggregation-side locals to make the plain-text matching intent explicit, and added comments in both places to document that rendering and aggregation are deliberately different concerns.

src/utils/__tests__/toasts-test.tsx
jeremy-davis-sonarsource: Aha, got it. Maybe a comment to explain this?

jeremy-davis-sonarsource (approved):
Noice!

One thing I noticed is that if there's a second toast on top, adding instances to the first one doesn't make it come back on top. I'm not sure we can do something about that though.

david-cho-lerat-sonarsource:
> Noice!
> 
> One thing I noticed is that if there's a second toast on top, adding instances to the first one doesn't make it come back on top. I'm not sure we can do something about that though.

doesn't seem supported by Sonner, I'm afraid. We'd need our own "hack" (dismiss old now-hidden toast, recreate it with proper count..), but that's a whole 'nother can of worms

src/utils/__tests__/toasts-test.tsx
david-cho-lerat-sonarsource: actually there's comments in trackToastId and resetToastTestState, so that should be enough I guess

PR #699, opened 2026-06-11
ECHOES-1337 ECHOES-1338 Align toast lifecycle defaults with notification guidelines

david-cho-lerat-sonarsource:
----
## Summary by Gitar

- **Refactored toast durations:**
  - Updated `Short`, `Medium`, and `Long` durations to 3s, 5s, and 8s respectively.
  - Set `Medium` as the default for regular toasts and `Long` as the default for toasts with actions.
- **Updated configuration:**
  - Reduced `EchoesProvider` default `toastsVisibleNb` from 5 to 3.
- **Improved type safety and documentation:**
  - Removed strict requirement for `infinite` duration on toasts with actions.
  - Clarified usage guidelines for `infinite` duration toasts in JSDoc and `Toast-stories`.
- **Updated testing:**
  - Adjusted existing tests to reflect new duration timings.
  - Added new test coverage for default auto-dismiss behavior on regular and action-based toasts.

### DIFF
--- a/src/utils/toasts.tsx
+++ b/src/utils/toasts.tsx
@@ -35,20 +35,22 @@ export { ToastVariety } from '~common/components/Toast';
  */
 export enum ToastDuration {
   /**
-   * Short duration for toasts with only 1 line of text (8 seconds).
+   * Short duration (3 seconds) for brief transient toasts when a shorter-than-default lifetime is
+   * intentional.
    */
   Short = 'short',
   /**
-   * Medium duration for toasts with 2 lines of text (16 seconds) - default duration.
+   * Medium duration (5 seconds). This is the default for regular toasts.
    */
   Medium = 'medium',
   /**
-   * Long duration for toasts with 3 lines of text (24 seconds).
+   * Long duration (8 seconds). This is the default for toasts with actions, or for regular toasts
+   * that intentionally need a bit more time.
    */
   Long = 'long',
   /**
-   * Infinite duration toast that remains visible until manually dismissed.
-   * Mandatory when the toast has actions.
+   * Infinite duration for toasts that must remain visible until the user dismisses them, such as
+   * critical or complex errors.
    */
   Infinite = 'infinite',
 }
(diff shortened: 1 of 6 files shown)


src/utils/toasts.tsx
```diff
@@ -35,20 +35,22 @@ export { ToastVariety } from '~common/components/Toast';
...
-   * Long duration for toasts with 3 lines of text (24 seconds).
+   * Long duration (8 seconds). This is the default for toasts with actions, or for regular toasts
+   * that intentionally need a bit more time.
    */
   Long = 'long',
   /**
-   * Infinite duration toast that remains visible until manually dismissed.
-   * Mandatory when the toast has actions.
+   * Infinite duration for toasts that must remain visible until the user dismisses them, such as
+   * critical or complex errors.
```
gregaubert, line 53: Are we really sure about the timing changes there?
* short - 3 seconds: I barely have the time to notice there is a toast that showed up, so no way I can read it...
* long - 8 seconds: default for toasts with actions... with actions really? there is no way we can expect someone to click on an action in toast that disappear after 8 seconds... The time to read the text, take a decision, move your mouse to click... 8 seconds there is no way. Toast with actions should be infinite and dismissable no?

src/utils/toasts.tsx
```diff
@@ -103,12 +101,12 @@ function toastFn(params: ToastParams, ref?: Ref<HTMLDivElement>): ToastId {
   const repeatedToastState = trackRepeatedToast(params);
   const repeatedToastId = repeatedToastState?.id;
   const repeatedToastCount = repeatedToastState?.count;
+  const hasActions = isDefined(toastProps.actions);
 
-  const durationValue = isDefined(toastProps.actions)
-    ? TOAST_DURATION_MAP[ToastDuration.Infinite]
-    : TOAST_DURATION_MAP[duration];
+  const resolvedDuration = duration ?? (hasActions ? ToastDuration.Long : ToastDuration.Medium);
```
gregaubert, line 106: I don't undersand and agree with this change... I even think having a timer on something that contains an action is against accessibility rules.

gregaubert (commented):
The code changes are looking fine. But I strongly disagree with the timing changes and the fact that toast with actions don't have the infinite duration anymore.

src/utils/toasts.tsx
david-cho-lerat-sonarsource: before this PR we had 8/16/24 for short/medium/long, but the ZH doc says 3/5/8 here: https://echoes.sonarsource.com/79d336d52/v/0/p/2062ab-notifications/t/fd8002b62b
even for toasts with action. But you're right to push back on that, that's something we should challenge with you @marciopmoreira6 & @2amelia, let's discuss

src/utils/toasts.tsx
david-cho-lerat-sonarsource: let's discuss with @marciopmoreira6 & @2amelia , maybe it's the ZH docs that need changing, not Echoes?

david-cho-lerat-sonarsource:
The ZH doc was not up to date

PR #701, opened 2026-06-11
ECHOES-1356 Add announcement modes to MessageCallout and MessageInline

david-cho-lerat-sonarsource:
----
## Summary by Gitar

- **New accessibility features:**
  - Added `announcementMode` prop to `MessageCallout` and `MessageInline` to enable `alert` or `status` live-region semantics.
- **Component updates:**
  - Integrated `LiveRegionAnnouncementMode` into both components to manage ARIA roles dynamically.
- **Testing and documentation:**
  - Added unit tests for default behaviors and live-region support.
  - Updated Storybook stories with new `announcementMode` controls and examples.
- **Refactoring:**
  - Corrected the `ARIA-label` description for the `MessageCallout` dismiss button in `i18n/keys.json` and component code.

### DIFF
--- a/src/components/messages/MessageInline.tsx
+++ b/src/components/messages/MessageInline.tsx
@@ -17,8 +17,10 @@
  * along with this program; if not, write to the Free Software Foundation,
  * Inc., 51 Franklin Street, Fifth Floor, Boston, MA  02110-1301, USA.
  */
+
 import styled from '@emotion/styled';
 import { forwardRef, PropsWithChildren, useMemo } from 'react';
+import { LiveRegionAnnouncementMode } from '~types/LiveRegionAnnouncementMode';
 import { MessageScreenReaderPrefix } from './MessageScreenReaderPrefix';
 import { MESSAGE_VARIETY_ICON } from './MessageStyles';
 import { MessageInlineSize, MessageVariety } from './MessageTypes';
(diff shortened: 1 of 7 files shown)


src/components/messages/MessageInline.tsx
```diff
@@ -17,15 +17,24 @@
...
 
 export interface MessageInlineProps {
+  /**
+   * Adds `status` or `alert` live-region semantics to the rendered message.
+   * Defaults to no explicit live-region semantics.
+   * Set `status` for contextual status updates or `alert` for urgent messages.
+   * Screen reader announcements are most reliable when an already-mounted
+   * message updates its content.
+   */
+  announcementMode?: `${MessageAnnouncementMode}`;
```
gregaubert: I'm a bit surprised that we are adding these possibilities to MessageCallout and MessageInline. I think when these two were initially created they were intended to answer different use cases.

Image

So I guess this is changing with this initiative and the zero height doc is updated too?

gregaubert (approved):
Looks good!

It might be worth adding the MessageInline and MessageCallout typescript documentation since we are touching these components.

src/components/messages/MessageInline.tsx
david-cho-lerat-sonarsource: I don't see changes in ZH specifically for these components, but in the UX initiative it says in ECHOES-1073 that MessageCallout and MessageInline are part of the "notification components", and the "Notifications" page on ZH, in the "Accessibility" tab, makes this a requirement

gregaubert: Let Amelia and Marcio know they should update ZH then

PR #708, opened 2026-06-17
ECHOES-1363 Implement Toolbar component

gregaubert:
----
## Summary by Gitar

- **New component:**
  - Implemented `Toolbar` component to support filtering, search, sorting, and dataset controls.
- **Component features:**
  - Added support for dynamic `filterTags` and a configurable "Clear filters" action.
  - Included contextual labels like `labelEmptyFilterTags` for empty states and `labelFilterTags` for applied counts.
- **Documentation and testing:**
  - Added comprehensive Storybook stories in `Toolbar-stories.tsx`.
  - Included unit and accessibility tests in `Toolbar-test.tsx`.
- **Localization:**
  - Added i18n keys for `toolbar.clear-all` and `toolbar.filter-tags-count` in `keys.json`.

### DIFF
--- a/src/components/toolbar/Toolbar.tsx
+++ b/src/components/toolbar/Toolbar.tsx
@@ -1,0 +1,241 @@
+/*
+ * Echoes React
+ * Copyright (C) 2023-2025 SonarSource Sàrl
+ * mailto:info AT sonarsource DOT com
+ *
+ * This program is free software; you can redistribute it and/or
+ * modify it under the terms of the GNU Lesser General Public
+ * License as published by the Free Software Foundation; either
+ * version 3 of the License, or (at your option) any later version.
+ *
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the GNU
+ * Lesser General Public License for more details.
+ *
+ * You should have received a copy of the GNU Lesser General Public License
+ * along with this program; if not, write to the Free Software Foundation,
+ * Inc., 51 Franklin Street, Fifth Floor, Boston, MA  02110-1301, USA.
+ */
+
+import styled from '@emotion/styled';
+import { ReactNode, Ref, useId } from 'react';
+import { FormattedMessage } from 'react-intl';
+import { isDefined } from '~common/helpers/types';
+import { cssVar } from '~utils/design-tokens';
+import { Button, ButtonSize, ButtonVariety } from '../buttons';
+import { Text } from '../typography';
+
+export interface ToolbarProps {
+  /**
+   * Accessible label for the toolbar group.
+   * Example: `"Issues filters and sorting"`.
+   */
+  ariaLabel?: string;
+
+  /**
+   * Optional CSS class name applied to the root element.
+   */
+  className?: string;
+
+  /**
+   * Content displayed at the trailing end of the top row (e.g. action buttons, item count).
+   * Automatically pushed to the right.
+   */
+  datasetControls?: ReactNode;
+
+  /**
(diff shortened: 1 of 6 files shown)


src/components/toolbar/Toolbar.tsx
```diff
@@ -0,0 +1,222 @@
...
+  flex: 0 0 auto;
+  min-width: 0;
+
+  // Only a full-width search input should stretch to fill the available space.
+  // SearchInput exposes its width mode through the data-width attribute.
+  &:has([data-width='full']) {
+    flex: 1 1 0;
+    max-width: 40%;
+    min-width: ${cssVar('sizes-form-field-medium')};
+  }
```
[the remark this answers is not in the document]
gregaubert: That's acceptable in this case

src/components/toolbar/Toolbar.tsx
```diff
@@ -0,0 +1,236 @@
...
+  display: flex;
+  align-items: center;
+  gap: ${cssVar('dimension-space-100')};
+  margin-left: auto;
+`;
+
+ToolbarDatasetControlsWrapper.displayName = 'ToolbarDatasetControlsWrapper';
+
+const ToolbarEmptyFilterLabel = styled(Text)`
+  padding-top: ${cssVar('dimension-space-75')};
```
jeremy-davis-sonarsource, line 237: Might be better to add padding-bottom as well, to preserve the full height

src/components/toolbar/Toolbar.tsx
```diff
@@ -0,0 +1,236 @@
...
+  const hasActiveTags = filterTags.length > 0;
+  const hasEmptyFilterTagsLabel = isDefined(labelEmptyFilterTags) && isDefined(filterControls);
+
+  return (
+    <ToolbarWrapper
+      {...restProps}
+      aria-label={ariaLabel}
+      className={className}
+      ref={ref}
+      role="group">
```
jeremy-davis-sonarsource, line 140: Why not a `toolbar`?
https://www.w3.org/WAI/ARIA/apg/patterns/toolbar/

edit: ok, I checked in detail, and it woul make the whole thing focusable, with keyboard navigation and everything... a bit over the top I guess. 
Leaving this here anyway to make sure we make it a conscious choice!

stories/toolbar/Toolbar-stories.tsx
```diff
@@ -0,0 +1,131 @@
...
+  return (
+    <Toolbar
+      {...props}
+      datasetControls={
+        <>
+          <Text>4 projects</Text>
+          <Button>Bulk action</Button>
+        </>
+      }
+      filterControls={
```
jeremy-davis-sonarsource, line 99: You should probably exclude all these props (not just the two highlighted!) from the story, since you're ignoring them.

src/components/toolbar/Toolbar.tsx
```diff
@@ -0,0 +1,236 @@
...
+  /**
+   * Filter controls (e.g., filter dropdowns) displayed after the search input.
+   */
+  filterControls?: ReactNode;
+
+  /**
+   * Active filter tags displayed in the applied-filters row.
+   * An empty array hides the tags and shows `labelEmptyFilterTags` instead (when configured).
+   */
+  filterTags?: ReactNode[];
```
jeremy-davis-sonarsource, line 57: Maybe specify we expect an array of <FilterTag /> ?

jeremy-davis-sonarsource (approved):
Looks great! Just a few things to iron out 👍

src/components/toolbar/Toolbar.tsx
gregaubert: I added a comment explaining why we keep group. Mainly because the keyboard navigation would conflict with the input and dropdown keyboard navigation.

PR #710, opened 2026-06-25
SC-51598 Set CheckboxGroup label spacing to 12px

marciopmoreira6:
## What does this PR do?

- Set CheckboxGroup label-to-options spacing to 12px when a group label is present.
- Add test coverage for the CheckboxGroup label spacing token.

Jira: https://sonarsource.atlassian.net/browse/SC-51598

## How to test?

1. Open the CheckboxGroup Valid story in Storybook.
2. Verify the visual gap between the group label and first checkbox is 12px.

## Validation

- yarn jest src/components/checkbox-group/__tests__/CheckboxGroup-test.tsx
- yarn prettier --check src/components/checkbox-group/CheckboxGroup.tsx src/components/checkbox-group/__tests__/CheckboxGroup-test.tsx
- yarn validate

----
## Summary by Gitar

- **Form component updates:**
  - Added `labelSpacing` prop to `FormField` to control the bottom margin of the label.
  - Introduced `FormFieldLabelSpacing` enum to support `default` and `large` spacing tokens.
- **Component styling:**
  - Updated `CheckboxGroup` and `RadioButtonGroup` to use `FormFieldLabelSpacing.Large` for better alignment.
- **Testing improvements:**
  - Added tests for `FormField` spacing variants and verified spacing in `CheckboxGroup` and `RadioButtonGroup`.

### DIFF
--- a/src/components/form/FormField.tsx
+++ b/src/components/form/FormField.tsx
@@ -22,9 +22,11 @@ import { type ComponentProps, type JSX, forwardRef } from 'react';
 import { TextNodeOptional } from '~types/utils';
 import { MessageInline, MessageInlineSize, MessageVariety } from '../messages';
 import { HelperText } from '../typography';
-import { FormFieldLabel, FormFieldLabelProps } from './FormFieldLabel';
+import { FormFieldLabel, FormFieldLabelProps, FormFieldLabelSpacing } from './FormFieldLabel';
 
 import { cssVar } from '~utils/design-tokens';
+
+export { FormFieldLabelSpacing };
 
 /**
  * Form fields wrap form controls and help create standardization between them.
@@ -136,11 +138,6 @@ export enum FormFieldWidth {
   Medium = 'medium',
   Large = 'large',
   Full = 'full',
-}
-
-export enum FormFieldLabelSpacing {
-  Default = 'default',
-  Large = 'large',
 }
 
 type WhiteListedProps = Pick<ComponentProps<'div'>, 'className'>;
--- a/src/components/form/FormField.tsx
+++ b/src/components/form/FormField.tsx
@@ -62,6 +62,7 @@ export const FormField = forwardRef<HTMLDivElement, FormFieldProps>((props, ref)
     isRequired = false,
     label,
     labelId,
+    labelSpacing = FormFieldLabelSpacing.Default,
     messageInvalid,
     messageValid,
     validation,
@@ -89,7 +90,8 @@ export const FormField = forwardRef<HTMLDivElement, FormFieldProps>((props, ref)
         htmlFor={controlId}
         id={labelId}
         isDisabled={isDisabled}
-        isRequired={isRequired}>
+        isRequired={isRequired}
+        spacing={labelSpacing}>
         {label}
       </FormFieldLabel>
       {children}
@@ -134,6 +136,11 @@ export enum FormFieldWidth {
   Medium = 'medium',
   Large = 'large',
   Full = 'full',
+}
+
+export enum FormFieldLabelSpacing {
+  Default = 'default',
+  Large = 'large',
 }
 
 type WhiteListedProps = Pick<ComponentProps<'div'>, 'className'>;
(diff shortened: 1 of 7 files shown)


gregaubert (changes requested):
With this change the RadioButtonGroup and CheckboxGroup are not aligned in style anymore, CheckboxGroup has more space between the label and the content of the group compared to the RadioButtonGroup. Do we really want that difference?

Also isn't this added space gonna be weird when these groups are used in forms? All the other form elements are gonna have a different spacing with their label. Should this change actually be applied to the FormField component to keep the consistency everywhere?

marciopmoreira6:
> With this change the RadioButtonGroup and CheckboxGroup are not aligned in style anymore, CheckboxGroup has more space between the label and the content of the group compared to the RadioButtonGroup. Do we really want that difference?
> 
> Also isn't this added space gonna be weird when these groups are used in forms? All the other form elements are gonna have a different spacing with their label. Should this change actually be applied to the FormField component to keep the consistency everywhere?

Good point. I moved the spacing into an internal FormField label spacing option and opted in both CheckboxGroup and RadioButtonGroup. This keeps both group components aligned at 12px while leaving other form controls on the default label spacing.

david-cho-lerat-sonarsource:
@marciopmoreira6 the review skill is saying this:

A few review notes from this pass:

Worth addressing — `src/components/form/FormField.tsx:141` and `src/components/form/FormFieldLabel.tsx:47`

The new spacing variant is defined twice: once as `FormFieldLabelSpacing`, and again as a separate `'default' | 'large'` union on `FormFieldLabelProps`. That gives this internal contract two sources of truth, so the next spacing variant has to be updated in both places and can drift.

Suggested fix: define the variant once and reuse it from both files. Cost: tiny internal cleanup, no consumer-facing impact.

Assumption: I’m assuming the 12px gap is intentionally limited to grouped controls. If the design goal is form-wide label consistency, this wants to be the default `FormField` spacing rather than an opt-in on just `CheckboxGroup` and `RadioButtonGroup`.

marciopmoreira6:
> @marciopmoreira6 the review skill is saying this:
> 
> A few review notes from this pass:
> 
> Worth addressing — `src/components/form/FormField.tsx:141` and `src/components/form/FormFieldLabel.tsx:47`
> 
> The new spacing variant is defined twice: once as `FormFieldLabelSpacing`, and again as a separate `'default' | 'large'` union on `FormFieldLabelProps`. That gives this internal contract two sources of truth, so the next spacing variant has to be updated in both places and can drift.
> 
> Suggested fix: define the variant once and reuse it from both files. Cost: tiny internal cleanup, no consumer-facing impact.
> 
> Assumption: I’m assuming the 12px gap is intentionally limited to grouped controls. If the design goal is form-wide label consistency, this wants to be the default `FormField` spacing rather than an opt-in on just `CheckboxGroup` and `RadioButtonGroup`.

Good catch. I moved the spacing variant into a single shared enum owned by FormFieldLabel and reused it from FormField, so there is no duplicated spacing contract. The 12px spacing remains intentionally opt-in for grouped controls only: CheckboxGroup and RadioButtonGroup.

src/components/form/FormField.tsx
```diff
@@ -192,6 +196,12 @@ export interface FormFieldProps extends ValidationProps, WhiteListedProps {
    * The ID of the label for the form field (optional).
    */
   labelId?: FormFieldLabelProps['id'];
+  /**
+   * Controls the spacing between the label and the form control.
+   *
+   * @internal
```
david-cho-lerat-sonarsource, line 202: `@internal` is more of an intent than enforcement. I’m a bit unsure about the API shape we’re creating here.

With `FormFieldLabelSpacing` + `labelSpacing`, this starts to look like a real reusable FormField API rather than just an internal implementation detail for CheckboxGroup / RadioButtonGroup.

If that’s the intent, I think it should be documented like the rest of the Echoes surface:
- enum members documented
- `@defaultValue` on the new defaulted props

If that’s *not* the intent, I wonder whether we should keep it a bit less public-looking, so consumers don’t start relying on a spacing knob we didn’t really mean to expose.

david-cho-lerat-sonarsource (changes requested):
should we expose FormFieldLabelSpacing + labelSpacing the way we're doing (@internal isn't preventing people from using them), and document them better, or should we make them truly internal?

src/components/form/FormField.tsx
david-cho-lerat-sonarsource: as discussed, it's better to leave this as-is now

david-cho-lerat-sonarsource (approved):

gregaubert (approved):

PR #711, opened 2026-06-25
SC-51613 Remove card body min-height

marciopmoreira6:
## Summary
- Remove the size-driven min-height from Card.Body
- Keep Card.Body padding controlled by card size
- Add a regression test for Card.Body min-height

## Test
- yarn jest src/components/card/__tests__/Card-test.tsx

----
## Summary by Gitar

- **Refactored component styles:**
  - Removed `min-height` and `max-height` variables from `CardHeaderStyled`, `RightContentStyled`, and `CardHeaderTitleButtonStyled`
  - Updated `CARD_HEADER_SIZE_STYLES` to rely on padding-based sizing instead of explicit height variables
- **Regression testing:**
  - Added tests to verify `min-height` is unset for `CardHeader`, `CardBody`, and collapsible header buttons
  - Added unit tests to ensure `CARD_HEADER_SIZE_STYLES` and `CARD_SIZE_STYLES` only use padding configurations

### DIFF
--- a/src/components/card/CardStyles.tsx
+++ b/src/components/card/CardStyles.tsx
@@ -84,6 +84,7 @@ export const CardHeaderTitleButtonStyled = styled(ButtonStyled)`
   display: flex;
   flex-direction: row;
   flex: 1;
+  /* ButtonStyled sets fixed dimensions; this header button sizes from padding instead. */
   height: auto;
   min-height: auto;
   padding: var(--card-header-padding);
--- a/src/components/card/CardStyles.tsx
+++ b/src/components/card/CardStyles.tsx
@@ -96,7 +96,6 @@ CardHeaderTitleButtonStyled.displayName = 'CardHeaderTitleButtonStyled';
 
 export const CardBodyStyled = styled.div<{ size: `${CardSize}`; insetContent: boolean }>`
   box-sizing: border-box;
-  min-height: var(--card-body-min-height);
   padding: ${(props) => (props.insetContent ? '0' : 'var(--card-padding)')};
   width: 100%;
 
@@ -126,15 +125,12 @@ export const CARD_HEADER_SIZE_STYLES = {
 
 export const CARD_SIZE_STYLES = {
   [CardSize.Large]: {
-    '--card-body-min-height': '108px',
     '--card-padding': cssVar('dimension-space-300'),
   },
   [CardSize.Medium]: {
-    '--card-body-min-height': '92px',
     '--card-padding': cssVar('dimension-space-200'),
   },
   [CardSize.Small]: {
-    '--card-body-min-height': '84px',
     '--card-padding': cssVar('dimension-space-150'),
   },
 };
--- a/src/components/card/CardStyles.tsx
+++ b/src/components/card/CardStyles.tsx
@@ -65,7 +65,6 @@ export const RightContentStyled = styled.div`
   align-items: center;
   display: flex;
   gap: ${cssVar('dimension-space-100')};
-  max-height: var(--card-header-heading-height);
 `;
 RightContentStyled.displayName = 'RightContentStyled';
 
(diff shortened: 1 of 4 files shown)


src/components/card/__tests__/Card-test.tsx
```diff
@@ -151,6 +182,16 @@ describe('Card components', () => {
...
 
+    it('does not apply a minimum height', () => {
+      render(
+        <CardRoot>
+          <CardBody>Content</CardBody>
+        </CardRoot>,
+      );
+
+      expect(window.getComputedStyle(screen.getByText('Content')).minHeight).toBe('');
+    });
```
gregaubert: This test is the same as a previous one, keep only one

src/components/card/CardStyles.tsx
```diff
@@ -86,7 +84,8 @@ export const CardHeaderTitleButtonStyled = styled(ButtonStyled)`
   display: flex;
   flex-direction: row;
   flex: 1;
-  min-height: var(--card-header-min-height);
+  height: auto;
+  min-height: auto;
```
gregaubert, line 89: Do we really need the auto value? Can't we just not specify the height? I think auto is the default value for the height anyway

marciopmoreira6: You're right that `auto` is the default for a plain element. In this case the styled component extends `ButtonStyled`, which already sets fixed `height` and `min-height` from `--button-height`. Without the explicit `auto` override, those button dimensions still apply and the header button does not hug the card header padding. I added a short code comment above the override to make that clear.

src/components/card/__tests__/Card-test.tsx
marciopmoreira6: Agreed. I consolidated the duplicate checks into one integrated test that asserts both the header and body do not apply a minimum height, so we keep the coverage without having two nearly identical tests.

david-cho-lerat-sonarsource:
@marciopmoreira6 the review skill is saying this:

Must-fix:  `src/components/card/__tests__/Card-test.tsx:231`

The consolidated `does not apply a minimum height to the header or body` test is still ineffective for the regression this PR is trying to guard against. Before this change, both removed implementations were `min-height: var(--card-*-min-height)`, and in jsdom `getComputedStyle(...).minHeight` stays `''` for unresolved `var(...)` values, so reintroducing the old rules would still keep this test green.

Suggested fix: assert on the style source of truth instead of computed style here, for example by verifying that `CARD_HEADER_SIZE_STYLES` / `CARD_SIZE_STYLES` no longer define the `--card-*-min-height` variables, or by asserting the emitted style rules on `CardHeaderStyled` / `CardBodyStyled` directly. Cost: small test-only follow-up, no runtime risk.

marciopmoreira6:
Handled in 0bbac48. I removed the `getComputedStyle(...).minHeight` regression check and extended the source-of-truth style-map assertion to cover both `CARD_HEADER_SIZE_STYLES` and `CARD_SIZE_STYLES`. Reintroducing `--card-header-min-height` or `--card-body-min-height` in those sizing maps will now fail the test instead of relying on jsdom resolving CSS variables.

marciopmoreira6:
Handled the two remaining Gitar suggestions in 5e73bf4: removed the unused `size` prop from `CardBodyStyled`/`CardBody` so it is not forwarded to the DOM, and renamed the internal `noPadding` flag to `disablePadding` to follow the boolean prop naming convention. I also added a Card body test assertion that the rendered body does not receive a `size` attribute.

src/components/card/CardStyles.tsx
```diff
@@ -108,33 +107,24 @@ CardBodyStyled.displayName = 'CardBodyStyled';
 
 export const CARD_HEADER_SIZE_STYLES = {
   [CardSize.Large]: {
-    '--card-header-heading-height': '1.875rem',
-    '--card-header-min-height': '58px',
-    '--card-header-padding': `${cssVar('dimension-space-200')} ${cssVar('dimension-space-300')}`,
+    '--card-header-padding': `${cssVar('dimension-space-250')} ${cssVar('dimension-space-300')}`,
```
david-cho-lerat-sonarsource, line 110: where is this 200 to 250 change for the header padding coming from? is this intentional in addition to removing the min-height? the PR only mentions removing min-height, but this will also have an impact on real product screens?

david-cho-lerat-sonarsource (changes requested):
if you confirm the 200 to 250 change was intentional then I have no further objections your honor

david-cho-lerat-sonarsource (approved):

PR #715, opened 2026-07-02
ECHOES-1408 Allow icon-less accordion child items in SidebarNavigation

david-cho-lerat-sonarsource:
NOTE: as we stand, with the sidebar still showing a "thin band" when collapsed, allowing no icons "breaks" the appearance, but this PR is the first in a series that will get rid of the "thin band" collapsed sidebar, and this will not be shipped on its own.

## Summary

- add `Layout.SidebarNavigation.AccordionItem.Item` for navigation rows inside sidebar accordion sections
- allow accordion child rows to omit their icon through the dedicated accordion child item API
- extract shared sidebar item logic and shared sidebar item base props
- update stories, tests, and exports for the accordion child item API

## Validation

- `yarn ts-check`
- `yarn jest src/components/layout/sidebar-navigation`
- targeted ESLint on the touched sidebar navigation files
- Sonar snippet analysis on the touched sidebar navigation files

----
## Summary by Gitar

- **New navigation components:**
  - Added `SidebarNavigationAccordionChildItem` for configurable child rows within accordions.
  - Introduced `SidebarNavigationBaseItem` to unify logic across all navigation item types.
  - Created a namespaced `SidebarNavigation.AccordionItem.Item` component for improved API ergonomics.
- **Improved functionality:**
  - Enabled icon-less rendering for accordion child items to support flexible UI layouts.
  - Added `ariaLabel` support across `SidebarNavigationItem`, `AccordionItem`, and child items.
  - Implemented `ref` forwarding for better accessibility and integration control.
- **Refactored structure:**
  - Migrated shared types to `SidebarNavigationTypes.ts` and consolidated styling.
  - Updated `SidebarNavigationItem` to use the base item implementation.
- **Testing & Documentation:**
  - Added comprehensive unit tests for the new child item component.
  - Updated `SidebarNavigation` stories to demonstrate the new accordion item API.

### DIFF
--- a/src/components/layout/index.ts
+++ b/src/components/layout/index.ts
@@ -22,6 +22,7 @@ import { Banner } from './banner';
 import { GlobalNavigation } from './global-navigation';
 import { ContentHeader, PageHeader } from './header';
 import { Layout as LayoutRoot } from './Layout';
+
 import {
   AsideLeft,
   BannerContainer,
@@ -30,9 +31,11 @@ import {
   PageFooter,
   PageGrid,
 } from './LayoutSlots';
+
 import { SidebarNavigation } from './sidebar-navigation';
 
 export { BannerVariety, type BannerProps } from './banner';
+
 export type {
   GlobalNavigationAccountProps,
   GlobalNavigationActionProps,
@@ -43,6 +46,7 @@ export type {
   GlobalNavigationPrimaryProps,
   GlobalNavigationProps,
 } from './global-navigation';
+
 export type {
   ContentHeaderProps,
   ContentHeaderTitleProps,
@@ -50,13 +54,17 @@ export type {
   PageHeaderProps,
   PageHeaderTitleProps,
 } from './header';
+
 export type { LayoutProps } from './Layout';
 export type { AsideProps, PageGridProps } from './LayoutSlots';
 export { AsideSize, PageWidth } from './LayoutTypes';
+
 export type {
+  SidebarNavigationAccordionChildItemProps,
   SidebarNavigationAccordionItemProps,
   SidebarNavigationGroupProps,
   SidebarNavigationHeaderProps,
+  SidebarNavigationItemBaseProps,
   SidebarNavigationItemProps,
   SidebarNavigationProps,
 } from './sidebar-navigation';
@@ -354,6 +362,7 @@ export const Layout = Object.assign(LayoutRoot, {
    * - {@link SidebarNavigation.Group} - Grouped navigation sections
    * - {@link SidebarNavigation.Item} - Individual navigation items
    * - {@link SidebarNavigation.AccordionItem} - Expandable sections
+   * - {@link SidebarNavigation.AccordionItem.Item} - Navigation rows within accordion sections
    * - {@link SidebarNavigation.Footer} - Fixed footer section
    *
    * ```tsx
--- a/src/components/layout/sidebar-navigation/SidebarNavigationAccordionItem.tsx
+++ b/src/components/layout/sidebar-navigation/SidebarNavigationAccordionItem.tsx
@@ -19,40 +19,48 @@
  */
 
 import styled from '@emotion/styled';
-import {
-  forwardRef,
-  ForwardRefExoticComponent,
-  ReactNode,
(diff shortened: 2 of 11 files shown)


src/components/layout/sidebar-navigation/SidebarNavigationTypes.ts
```diff
@@ -0,0 +1,83 @@
...
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the GNU
+ * Lesser General Public License for more details.
+ *
+ * You should have received a copy of the GNU Lesser General Public License
+ * along with this program; if not, write to the Free Software Foundation,
+ * Inc., 51 Franklin Street, Fifth Floor, Boston, MA  02110-1301, USA.
+ */
+
+import { ForwardRefExoticComponent, MouseEventHandler, ReactNode, Ref, RefAttributes } from 'react';
```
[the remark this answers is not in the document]
david-cho-lerat-sonarsource: Echoes is already moving to the React 19 `ref`-as-prop pattern for new code. The repo guidance in `.gitar/review/component-patterns.md` explicitly says not to introduce new `forwardRef` wrappers, and existing components such as `DismissButton` and `FilterTag` already follow that pattern.

This change matches that guidance, and the ref behavior is covered by the passing sidebar navigation tests and `yarn ts-check`, so I'm resolving these duplicated `forwardRef` suggestions.

src/components/layout/sidebar-navigation/SidebarNavigationAccordionItem.tsx
```diff
@@ -19,40 +19,53 @@
...
-  ReactNode,
-  useCallback,
-  useEffect,
-  useId,
-  useRef,
-  useState,
-} from 'react';
+
+import { ReactNode, Ref, useCallback, useEffect, useId, useRef, useState } from 'react';
+
```
gregaubert, line 24: Some weird spaces got added between the imports, we might be missing some formatting rules there

src/components/layout/index.ts
```diff
@@ -354,6 +362,7 @@ export const Layout = Object.assign(LayoutRoot, {
    * - {@link SidebarNavigation.Group} - Grouped navigation sections
    * - {@link SidebarNavigation.Item} - Individual navigation items
    * - {@link SidebarNavigation.AccordionItem} - Expandable sections
+   * - {@link SidebarNavigation.AccordionItem.Item} - Navigation rows within accordion sections
```
gregaubert, line 365: If I understand correctly now we have a new AccordionItem specifc item that we didn't have before and we should use as the content of the AccordionItem right?

And the only difference between the AccordionChildItem and the SidebarNavigationItem we where using before is the optional Icon for AccordionChildItem? Or are there other differences?

It seems like a lot of changes for making a prop optional. Couldn't we just make it optional on `SidebarNavigationItem` and that's it?

david-cho-lerat-sonarsource: well we don't want people to be able to not use icons on `SidebarNavigationItem`, they should remain mandatory there, so that's the whole point of this

gregaubert: But they could use an AccordionItem instead of a SidebarNavigationItem and not provide an Icon right? If they want they can, even with this whole implem.

david-cho-lerat-sonarsource: yeah but it's obvious from the name that they shouldn't do that. I know it's not 100% "safe", but at least we're nudging them

gregaubert: Yeah I'm mainly wondering if the nudge is worth that many changes.

Also I was kind of under the impression that we wanted to have mantadory icons in the sidebar because of the mechanism to collapse that shows only icons, but since we are getting rid of this and the sidebar can only be visible full width we don't really need it to be mandatory anymore no?

david-cho-lerat-sonarsource: Marcio still wanted the mandatory icons except in accordion items, where they could be added but defaulted to no icon

gregaubert (approved):

PR #738, opened 2026-07-21
ECHOES-1433 amelia/add-qg-indicator-tokens

2amelia:
Adds Quality Gate indicator tokens to Echoes so QG states can be styled consistently through the design token system instead of ad hoc values or bending other tokens to fit it.


design-tokens/tokens/component/dark.json
```diff
@@ -679,6 +679,48 @@
...
         }
       }
+    },
+    "quality-gate-indicator": {
+      "colors": {
+        "passed": {
+          "icon": {
+            "$type": "color",
+            "$value": "{echoes.color.roles.success.light}"
+          },
```
jeremy-davis-sonarsource: NP: they're alphabetical, can you put it in its rightful place?

jeremy-davis-sonarsource (approved):
I think Gitar's comment about using palette values on one side and roles on the other is a valid question.
Otherwise just a nitpick!

